PR #32, opened 2023-06-29
tmp: start doc

chaen:


docs/README.md
```diff
@@ -0,0 +1,161 @@
+## Run a full diracx demo
+
+This will allow you to run a demo setup.
+
+The code changes will be reflected in the demo.
+
+Requirement: docker, internet
+
+```bash
+# Clone the diracx repository
```
cserf, line 10: Clone your fork ?

docs/README.md
```diff
@@ -0,0 +1,161 @@
...
+git clone git@github.com:<yourfork>/diracx.git
+
+# Clone the diracx-chart repository
+git clone git@github.com:<yourfork>/diracx-charts.git
+
+# Run the
+diracx-charts/run_demo.sh diracx/
+```
```
fstagni, line 18:
```suggestion

0. fork, in GitHub, the https://github.com/DIRACGrid/diracx/ and https://github.com/DIRACGrid/diracx-charts/ repositories.

```bash
git clone git@github.com:<yourfork>/diracx.git
cd diracx
git remote add upstream git@github.com:DIRACGrid/diracx.git
git fetch upstream

# Clone the diracx-chart repository
git clone git@github.com:<yourfork>/diracx-charts.git
cd diracx-charts
git remote add upstream git@github.com:DIRACGrid/diracx-charts.git
git fetch upstream


# Run the
diracx-charts/run_demo.sh diracx/
```
```

File: docs/README.md [hunk 3]
```diff
@@ -0,0 +1,161 @@
...
+
+## Create the dev environment
+
+This will help you setup a dev environment to run the unit tests
+
+Requirements: conda, git
+
+```bash
+# Clone the diracx repository
+git clone git@github.com:DIRACGrid/diracx.git
```
fstagni, line 28:
```suggestion
```

docs/README.md
simon-mazenoux, line 18: Make the "login" process more explicit, by specifying that the email / password should be added in the identity provider page and not in the Swagger interface.

docs/README.md
aldbr: What if I just want to try the demo? (and not forking the whole project)

docs/README.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+### Run a local instance of diracx
+
+This only runs the diracx server, not any dependency like external IdP.
+
+
+```bash
+# In the `diracx` folder
+./run_local.sh
```
ruslan33, line 65: In my case I should correct following in the script: 
tmp_dir="/Users/ruslan/BelleII/DIRACX/dev/tmp" #explicitly name tmp dir
export DIRACX_CONFIG_BACKEND_URL="git+file://${tmp_dir}/cs_store/initialRepo" #add this env var

docs/README.md
```diff
@@ -0,0 +1,161 @@
...
+pre-commit install
+```
+
+### Run the test
+
+Run the unit tests
+
+```bash
+# In the `diracx` folder
+pytest
```
fstagni, line 53:
```suggestion
`pytest` for finding and running all tests. `pytest --regenerate-client tests/db` for running a subset of them. 
```

docs/README.md
```diff
@@ -0,0 +1,161 @@
...
+mypy .
+pre-commit run --all-files
```
fstagni, line 55:
```suggestion
`mypy .`
`pre-commit run --all-files`
```

docs/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+Modify the package's `setup.cfg`
+
+```ini
+[options.entry_points]
+diracx.services =
+	parking = diracx.routers.parking:router
+```
+
+This will prefix the routes with `/parking/` and mark them with the `"parking"` tag in the OpenAPI spec.
```
chaen, line 140: Add a line that it requires to reinstall diracx with pip to update this list

chaen:
Add comments regarding the DIRAC refactoring

docs/README.md
chaen: I don't think this is necessary, it makes the instructions longer for no added value really. Plus, if they want to run a demo, they don't need to fork it.

docs/README.md
chaen: why ? They are in a code block

src/diracx/db/__init__.py
```diff
@@ -1,4 +1,6 @@
-__all__ = ("AuthDB", "JobDB")
+__all__ = ("AuthDB", "JobDB", "DummyDB")
```
chrisburr, line 1:
```suggestion
__all__ = ("AuthDB", "JobDB")
```

docs/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+```python
+from datetime import datetime, timedelta, timezone
+
+import pytest
+```
+
+### Databases
+
+All SQL database interaction should use SQLAlchemy core and not the ORM to avoid performance issues.
```
chrisburr, line 178: This is already mentioned higher up

```suggestion
```

src/diracx/db/__init__.py
```diff
@@ -1,4 +1,6 @@
...
 
 from .auth.db import AuthDB
 from .jobs.db import JobDB
+
+# from .dummy.db import DummyDB
```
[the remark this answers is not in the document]
chaen: I am actually undecided.
That is what you would do for a proper DB. Agreed, this is just an example (which probably should be moved as an extension-demo, @cserf started this btw), but the aim is still to show how to do. 
So I am not sure we should remove it. Maybe from the `__all__`, but not the import

src/diracx/db/dummy/db.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
chaen: indeed, see my comment above

fstagni:
Just for discussion: we have in DIRAC what IMHO is a very nice documentation setup with readthedocs. Personally, I would replicate that here.

chaen:
Which documentation are you referring to exactly ? 
The problem is that setting up `DIRAC` is so special (a bit less nowadays) that it required plenty of documentation. But I don't think I have ever seen any other project going as far as telling you to click the *clone* button and what not.

src/diracx/db/__init__.py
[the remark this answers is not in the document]
chaen: they are indeed much more meaningful but harder to follow. If you try to search for `AuthDB` you'll find plenty of things which are unrelated to how you actually create and integrate the class. Looking for `DummyDB` reduces that noise.

fstagni:
> Which documentation are you referring to exactly ? 

https://dirac.readthedocs.io/en/latest/

chaen:
> > Which documentation are you referring to exactly ?
> 
> https://dirac.readthedocs.io/en/latest/

I am probably a bit confused then. Although I think the `DIRAC` doc is a mess, we indeed should have a very comprehensive doc, probably on readthedocs. But I thought you were referring to installation instructions specifically ?

fstagni:
Sorry, maybe I was not very clear. I think that:

- installation instructions/quick how-tos for developers etc. can/should be in the local README.md file. 
- user guide and administrator guide should be in readthedocs. Also code documentation can be there.

The DIRAC readthedocs is large and at times confusing because of how "special" DIRAC is. Content apart, I like its setup, which I would replicate for diracx.

src/diracx/db/__init__.py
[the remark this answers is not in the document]
chrisburr: At the moment the code is invalid as `__all__` contains something which is undefined. If you want to keep the commented import then the entry in `__all__` should also be commented.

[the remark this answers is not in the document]
chaen: OK so to conclude: I'll just comment it out for now, but will move all this promptly in the extension example

chaen:
OK, I understand better. Yes we can do something like that

chrisburr (approved):

PR #35, opened 2023-06-30
Replace print statements with proper logging

chrisburr:


src/diracx/routers/job_manager/__init__.py
```diff
@@ -186,7 +188,7 @@ async def submit_bulk_jobs(
             user_info.vo,
         )
 
-        print(
+        logging.debug(
             f'Job added to the JobDB", "{job_id} for {fixme_ownerDN}/{fixme_ownerGroup}'
```
aldbr, line 192:
```suggestion
        logger.debug(
            "Job added to the JobDB: %s for %s/%s" % (job_id, fixme_ownerDN, fixme_ownerGroup)
```

You are using the printf formatting style over the f-strings style for optimization purposes, aren't you?

chrisburr: This still needs discussion and personally I've not settled yet on a decision of what the best option is. This PR mostly exists so we can include some kind of logging tomorrow/Wednesday and refactor later if needed.

There are a couple of advantages of using the `logger.level(message, *args)` style:

* When running with a JSON log formatter you have separate access to the message template and arguments so you can aggregate based on the template in elasticsearch rather than the formatted string.
* Evaluating the string is deferred until it's absolutely needed so if you're running with debug messages disabled you don't pay the cost of building strings you don't need. (It's normally cheap, but not always.)
* It's easier to automatically refactor to f-strings than it is to convert f-strings to this style.

Issue #13, opened 2023-06-27
Implement support for job sandboxes


chrisburr:
The aim of this should be to be able to submit a job to DiracX with an input sandbox and be able to run it with the existing DIRAC infrastructure.

There are two parts of this as uploaded sandboxes need to be available via both DISET and DiracX.

### How sandboxes currently work?

* Files are uploaded with DISET and the path is `SE:ProductionSandboxSE:/Sandbox/u/username.group/hash.tar.gz` where `ProductionSandboxSE` comes from the CS (`LocalSEName`) and `/Sandbox/` is hard coded.
* Data is stored on the service's local disk.
* Data is deduplicated for a given user+group using the file's MD5 hash.
* Files are downloaded with DISET from local storage.

### How will sandboxes work in DiracX?

* Data will be stored on S3-compatible storage (default is minio from the helm chart)
* Sandbox paths will be `SE:ProductionSandboxSE:/S3/u/username.group/hash.tar.gz`
* Doing `POST /jobs/sandbox` returns a payload indicating the presigned URL which is used to send the data. The body of the post request includes the SHA-256 of the payload and the payload size.
* If the data already exists a JSON blob is returned containing the sandbox ID which should be included in the JDL.
* If the data doesn't exist, a JSON blob including S3 pre-signed URL is returned. This pre-signed URL should use the [`x-amz-content-sha256`](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html) header to make the storage verify the hash of the sandbox and limit the `Content-Lenth`.
* Data is downloaded by doing `GET /jobs/sandbox/SE:ProductionSandboxSE:/S3/u/username.group/hash.tar.gz` with a HTTP temporary redirect to a pre-signed URL.

Tasks to implement this in DiracX:

- [ ] Add minio to helm chart
- [ ] Make a DiracX DB interface with the `SandboxMetadataDB`
- [ ] Add a sandbox upload route to the jobs router
- [ ] Add a sandbox download route the jobs router

For assigning sandboxes to job IDs (implementation tbd):

- [ ] Input sandboxes are handled by the `JobSanity` executor
- [ ] Output sandboxes

### Migration

The exisiting `SandboxStoreHandler` need to be modified to:

- [ ] Proxy data from S3 if the path starts with `/S3/`.
- [ ] The existing `SandboxStoreHandler` should have a flag (`UseS3Backend`) added to make it upload to S3 instead of the local disk and return the appropriate path.

[...]

PR #42, opened 2023-07-04
Partially adding sandbox metadata db

fstagni:
This PR adds the basic functionalities of SandboxMetadataDB. Partially addresses https://github.com/DIRACGrid/diracx/issues/13 . Few notes:

- the DB schema added here is what is defined in https://github.com/DIRACGrid/DIRAC/pull/6566 (so, what is in DIRAC v8.1 but excluding OwnerDN).
- Only insertion of a new SB is added, no other functionalities (deletion is tentatively added)
- There are discussions going on in https://github.com/DIRACGrid/diracx/discussions/16 that might lead to a simplification of DIRAC's Sandbox (specifically if external Sandbox SEs are not supported) so I would wait before implementing further.

### DIFF
--- a/src/diracx/db/sandbox_metadata/db.py
+++ b/src/diracx/db/sandbox_metadata/db.py
@@ -1,0 +1,80 @@
+""" SandboxMetadataDB frontend
+"""
+
+from __future__ import annotations
+
+import datetime
+
+import sqlalchemy
+
+from ..utils import BaseDB
+from .schema import Base as SandboxMetadataDBBase
+from .schema import sb_Owners, sb_SandBoxes
+
+
+class SandboxMetadataDB(BaseDB):
+    # This needs to be here for the BaseDB to create the engine
+    metadata = SandboxMetadataDBBase.metadata
+
+    async def _get_put_owner(self, owner: str, owner_group: str) -> int:
+        """adds a new owner/ownerGroup pairs, while returning their ID if already existing
+
+        Args:
+            owner (str): user name
+            owner_group (str): group of the owner
+        """
+        stmt = sqlalchemy.select(sb_Owners.OwnerID).where(
+            sb_Owners.Owner == owner, sb_Owners.OwnerGroup == owner_group
+        )
+        result = await self.conn.execute(stmt)
+        if owner_id := result.scalar_one_or_none():
+            return owner_id
+
+        stmt = sqlalchemy.insert(sb_Owners).values(Owner=owner, OwnerGroup=owner_group)
+        result = await self.conn.execute(stmt)
+        return result.lastrowid
+
+    async def insert(
+        self, owner: str, owner_group: str, sb_SE: str, se_PFN: str, size: int = 0
+    ) -> tuple[int, bool]:
+        """inserts a new sandbox in SandboxMetadataDB
+        this is "equivalent" of DIRAC registerAndGetSandbox
+
+        Args:
+            owner (str): user name_
+            owner_group (str): groupd of the owner
+            sb_SE (str): _description_
+            sb_PFN (str): _description_
+            size (int, optional): _description_. Defaults to 0.
+        """
+        owner_id = await self._get_put_owner(owner, owner_group)
+        stmt = sqlalchemy.insert(sb_SandBoxes).values(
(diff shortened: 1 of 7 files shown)


src/diracx/db/sandbox_metadata/db.py
```diff
@@ -0,0 +1,86 @@
...
+            owner (str): _description_
+            owner_group (str): _description_
+        """
+        stmt = sqlalchemy.select(sb_Owners.OwnerID).where(
+            sb_Owners.Owner == owner, sb_Owners.OwnerGroup == owner_group
+        )
+        result = await self.conn.execute(stmt)
+        res = result.fetchall()
+        if res:
+            return res[0][0]
```
chrisburr, line 29: I think `scalar_one_or_none` is the idiomatic way of doing this:

```suggestion
        if owner_id := result.scalar_one_or_none():
            return owner_id
```

(with the benifit that if the impossible happens and multiple results are returned we get an exception)

src/diracx/db/sandbox_metadata/schema.py
```diff
@@ -0,0 +1,54 @@
...
+
+class sb_SandBoxes(Base):
+    __tablename__ = "sb_SandBoxes"
+    SBId = Column(Integer, autoincrement=True)
+    OwnerId = Column(Integer)
+    SEName = Column(String(64))
+    SEPFN = Column(String(512))
+    Bytes = Column(BigInteger)
+    RegistrationTime = Column(DateTime, default=datetime.datetime.utcnow())
+    LastAccessTime = Column(DateTime, default=datetime.datetime.utcnow())
```
chrisburr, line 36:
```suggestion
    RegistrationTime = DateNowColumn()
    LastAccessTime = DateNowColumn()
```

tests/db/test_sandboxMetadataDB.py
```diff
@@ -0,0 +1,80 @@
...
+        result = await sandbox_metadata_db.insert(
+            "owner",
+            "owner_group",
+            "sbSE",
+            "sbPFN",
+            123,
+        )
+        assert result == (1, True)
+
+        result = await sandbox_metadata_db.delete(1)
```
chrisburr, line 79: According to the type hint this should be:
```suggestion
        result = await sandbox_metadata_db.delete([1])
```

src/diracx/db/sandbox_metadata/db.py
```diff
@@ -0,0 +1,86 @@
...
+            if res:
+                sb_ID = res[0][0]
+                stmt = (
+                    sqlalchemy.update(sb_SandBoxes)  # type: ignore
+                    .where(sb_SandBoxes.SBId == sb_ID)
+                    .values(LastAccessTime=datetime.datetime.utcnow())
+                )
+                await self.conn.execute(stmt)
+                return sb_ID, False
+            return 0, False
```
chrisburr, line 73: It's not obvious to me how this line is supposed to be reachable?

If it isn't reachable, the return type of the function can be simplified from `tuple[int, bool]` to `int`.

src/diracx/db/sandbox_metadata/db.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Args:
+            owner (str): _description_
+            owner_group (str): _description_
+            sb_SE (str): _description_
+            sb_PFN (str): _description_
+            size (int, optional): _description_. Defaults to 0.
+        """
+        owner_id = await self._get_put_owner(owner, owner_group)
+        stmt = sqlalchemy.insert(sb_SandBoxes).values(
```
chrisburr, line 49: I think you can get rid of the `# type: ignore` comments if you use:
```suggestion
        stmt: sqlalchemy.Executable = sqlalchemy.insert(sb_SandBoxes).values(
```

An aside (which hopefully isn't needed at all in this case), it's probably better to stick to `# type: ignore[no-redef]` rather then the catch all `# type: ignore` comment so it's self documenting.

src/diracx/db/sandbox_metadata/db.py
fstagni: It seems reachable to me.

chrisburr: How? Either the row already exists or it gets added (and we're within a single DB transaction so we can't have a race).

Issue #23, opened 2023-06-29
Implement JobPolicy 


chaen:
- [ ] Update the CS model
- [ ] Factorize JobPolicy in vanilla DIRAC
- [ ] Add `JobPolicy` to `/jobs` router

panta-123:
Hi @chaen , Is there some documentation or instrcution for developers on how to contribute ?

aldbr:
Closing the issue as we now have access policies in diracx.

PR #43, opened 2023-07-05
feat: adding bulk jdl submission and job policy checking

simon-mazenoux:
Adding bulk submission of jdls to the job submission endpoint, and implement job policy checking (solves  Implement JobPolicy #23 )

(Tests need to be relaunched once https://github.com/DIRACGrid/DIRAC/pull/7092 get merged)

### DIFF
--- a/src/diracx/routers/job_manager/__init__.py
+++ b/src/diracx/routers/job_manager/__init__.py
@@ -174,7 +174,7 @@ async def submit_bulk_jobs(
     # TODO: make the max number of jobs configurable in the CS
     if len(jobDescList) > MAX_PARAMETRIC_JOBS:
         raise HTTPException(
-            status_code=HTTPStatus.UNAUTHORIZED,
+            status_code=HTTPStatus.BAD_REQUEST,
             detail=f"Normal user cannot submit more than {MAX_PARAMETRIC_JOBS} jobs at once",
         )
 
--- a/tests/routers/test_job_manager.py
+++ b/tests/routers/test_job_manager.py
@@ -182,7 +182,7 @@ def test_user_cannot_submit_parametric_jdl_greater_than_max_parametric_jobs(
     """Test that a user cannot submit a parametric JDL greater than the max parametric jobs"""
     job_definitions = [TEST_LARGE_PARAMETRIC_JDL]
     res = normal_user_client.post("/jobs/", json=job_definitions)
-    assert res.status_code == HTTPStatus.UNAUTHORIZED, res.json()
+    assert res.status_code == HTTPStatus.BAD_REQUEST, res.json()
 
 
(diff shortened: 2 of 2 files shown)


src/diracx/routers/job_manager/__init__.py
```diff
@@ -114,54 +116,77 @@ async def submit_bulk_jobs(
...
+            self.userGroup = user_info.dirac_group
+            self.userProperties = user_info.properties
+            self.jobDB = None
+            self.allInfo = allInfo
+            self._permissions: dict[str, bool] = {}
+            self._getUserJobPolicy()
+
+    # Check job submission permission
+    result = DiracxJobPolicy(user_info).getJobPolicy()
+    print(result)
```
aldbr, line 131:
```suggestion

```

fstagni:
This effectively depends from https://github.com/DIRACGrid/diracx/pull/67 (which in turn depends from https://github.com/DIRACGrid/DIRAC/pull/7143). I convert to draft until the 2 above are sorted.

src/diracx/routers/job_manager/__init__.py
```diff
@@ -104,64 +105,78 @@ async def submit_bulk_jobs(
...
+                    detail="You cannot submit parametric jobs in a bulk fashion",
+                )
+
+        jobDescList = job_definitions
+        parametricJob = True
+
+    # TODO: make the max number of jobs configurable in the CS
+    if len(jobDescList) > MAX_PARAMETRIC_JOBS:
+        raise HTTPException(
+            status_code=HTTPStatus.UNAUTHORIZED,
```
chrisburr, line 177: This should be HTTP 400 Bad Request

src/diracx/routers/job_manager/__init__.py
```diff
@@ -104,64 +105,78 @@ async def submit_bulk_jobs(
...
-        result = generateParametricJobs(jobClassAd)
+    class DiracxJobPolicy(JobPolicy):
+        def __init__(self, user_info: UserInfo, allInfo: bool = True):
+            self.userName = user_info.preferred_username
+            self.userGroup = user_info.dirac_group
+            self.userProperties = user_info.properties
+            self.jobDB = None
+            self.allInfo = allInfo
+            self._permissions: dict[str, bool] = {}
+            self._getUserJobPolicy()
```
chrisburr, line 123: Can we define this outside the method rather than creating a new class every time the API route is called?

simon-mazenoux: This was more a kind of proof of concept while waiting for a decision about attributing directly the properties during the token creation. In any case, this will need to be refactored because JobPolicy is also used in endpoinds such as kill, delete or reschedule

PR #44, opened 2023-07-05
use realpath to add the yaml file when generating the CS

hamar:
Otherwise we get exceptions like 

```
ValueError: Absolute path PosixPath('/var/folders/50/d5nfyk3x1qjdp7fwrnwjwtmm0000gn/T/tmp.Yfn6F3fz/cs_store/initialRepo/default.yml') is not in git repository at 
'/private/var/folders/50/d5nfyk3x1qjdp7fwrnwjwtmm0000gn/T/tmp.Yfn6F3fz/cs_store/initialRepo'
```

### DIFF
--- a/src/diracx/cli/internal.py
+++ b/src/diracx/cli/internal.py
@@ -58,13 +58,13 @@ def generate_cs(
         DIRAC=DIRACConfig(),
         Operations={"Defaults": OperationsConfig()},
     )
-
     repo = git.Repo.init(repo_path, initial_branch="master")
     yaml_path = repo_path / "default.yml"
     typer.echo(f"Writing configuration to {yaml_path}", err=True)
     config_data = json.loads(config.json(exclude_unset=True))
     yaml_path.write_text(yaml.safe_dump(config_data))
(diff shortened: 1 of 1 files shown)


src/diracx/cli/internal.py
```diff
@@ -104,7 +104,8 @@ def add_user(
     typer.echo(f"Writing back configuration to {yaml_path}", err=True)
     config_data = json.loads(config.json(exclude_unset=True))
     yaml_path.write_text(yaml.safe_dump(config_data))
-    repo.index.add([yaml_path])
+    # MacOS adds a "/private" in front, so we need to resolve the path when adding
+    repo.index.add([yaml_path.resolve()])
```
chrisburr, line 108: Maybe changing line 103 to `yaml_path = "default.yml"` would be preferable? (And the same in `generate_cs`)

chaen:
I can't reproduce anymore, and I suspect that `pytest` generates the `realpath` now. So this fix is not needed anymore.
We can use relative path (although the doc explicitely states that absolue path are supported)

PR #53, opened 2023-07-28
feat: adding JobLoggingDB

fstagni:


chaen:
I think @simon-mazenoux started this too

PR #61, opened 2023-08-14
Style: aligned black with ruff settings, added isort

fstagni:


fstagni:
The issue that `pytest` spots is described in https://github.com/tiangolo/typer/issues/533. There's a connected PR, but not merged yet.

fstagni:
This PR is ready to be merged. The only question I have is if ruff/pyupgrade should be forbidden to run on the auto-generated client code.

src/diracx/cli/__init__.py
```diff
@@ -22,8 +22,8 @@
 @app.async_command()
 async def login(
     vo: str,
-    group: Optional[str] = None,
-    property: Optional[list[str]] = Option(
+    group: Optional[str] = None,  # noqa: UP007
```
chrisburr, line 25: Shall we do this at the file level with a comment explaining why?

```
# ruff: noqa: F841
```

https://beta.ruff.rs/docs/configuration/#error-suppression

chrisburr:
> The only question I have is if ruff/pyupgrade should be forbidden to run on the auto-generated client code.

I think we definitely shouldn't, we just risk introducing bugs.

fstagni:
Superseeded by https://github.com/DIRACGrid/diracx/pull/74

Issue #63, opened 2023-08-24
Implement TaskQueueDB


simon-mazenoux:
* [x] Implement the schema
* [x] Implement the TaskQueueDB class
* [ ] Implement the relevent endpoints

ryuwd:
We don't need endpoints for TaskQueueDB, please close

Issue #22, opened 2023-06-29
JobManagerHandler conversion


chaen:
Continue the implementation of the `/jobs` route.
Missing methods are:

- [ ] getMaxParametricJobs: no need to implement in the router. The legacy adapter should query the CS directly.
- [ ] submitJob(self, jobDesc): started, needs completion (JobParametersDB, JobLoggingDB, etc)
- [ ] confirmBulkSubmission(self, jobIDs) : needs larger discussion about ParametricJobSubmission
- [ ] rescheduleJob(self, jobIDs):  `POST` to `/jobs/{job_id}/reschedule` and `/jobs/reschedule` for bulk
- [ ] removeJob(self, jobIDs): Do we need both `removeJob` and `deleteJob` ?  (If yes, can `removeJob` be a parameter of `deleteJob`)
- [ ] deleteJob(self, jobIDs): see above
- [ ] killJob(self, jobIDs): `POST` to  `/jobs/{job_id}/kill` and `/jobs/kill` for bulk (see [1])
- [ ] resetJob(self, jobIDs): Could this be a parameter to `rescheduleJob`?

[1] We could think of having `killJob` and `rescheduleJob` react to `setStatus(kill)` or `setStatus(reschedule)`. We do not because in general, it is forbidden to set an arbitrary status. `Kill` and `Reschedule` are exceptions, that's why they have a dedicated API.

ryuwd:
In addition to a lot of refactoring: 

-  getMaxParametricJobs: no need to implement in the router. The legacy adapter should query the CS directly.
  - Left this alone (TBD)
- submitJob(self, jobDesc): started, needs completion (JobParametersDB, JobLoggingDB, etc)
  - Left this alone (TBD)
- confirmBulkSubmission(self, jobIDs) : needs larger discussion about ParametricJobSubmission
   - Left this alone (TBD)
- rescheduleJob(self, jobIDs): POST to /jobs/{job_id}/reschedule and /jobs/reschedule for bulk
  - TBD
- removeJob(self, jobIDs): Do we need both removeJob and deleteJob ? (If yes, can removeJob be a parameter of deleteJob)
  - This has become a DELETE operation
- deleteJob(self, jobIDs): see above
  - This has become a DELETE operation
- killJob(self, jobIDs): POST to /jobs/{job_id}/kill and /jobs/kill for bulk (see [1])
[...]

PR #64, opened 2023-08-24
Implement kill, delete and remove endpoints

simon-mazenoux:
Implement kill, delete and remove endpoints (see #22, #63)


simon-mazenoux:
In this PR, and unlike what is done in DIRAC, I used the ON DELETE CASCADE constraint. This change would therefore require an ALTER statement on some tables of the DB. Should we do this or duplicate the logic of what is currently done in DIRAC ?

chrisburr:
> In this PR, and unlike what is done in DIRAC, I used the ON DELETE CASCADE constraint. This change would therefore require an ALTER statement on some tables of the DB. Should we do this or duplicate the logic of what is currently done in DIRAC ?

This seems very sensible to me, would any changes be needed in DIRAC v8.1 to make it tollerant to this? We should probably discuss it in the BiLD, can you open an issue in vanilla DIRAC?

simon-mazenoux:
> > In this PR, and unlike what is done in DIRAC, I used the ON DELETE CASCADE constraint. This change would therefore require an ALTER statement on some tables of the DB. Should we do this or duplicate the logic of what is currently done in DIRAC ?
> 
> This seems very sensible to me, would any changes be needed in DIRAC v8.1 to make it tollerant to this? We should probably discuss it in the BiLD, can you open an issue in vanilla 

I think DIRAC is already tolerant:  by adding the `ON DELETE CASCADE` constraint, we would just avoid having to delete manually every foreign key reference to the row we want to delete. Therefore, I don't really see how it could affect the codebase besides not raising an error anymore if a foreign key is not deleted before deleting the primary key.

src/diracx/db/jobs/schema.py
```diff
@@ -198,3 +210,99 @@ class LoggingInfo(JobLoggingDBBase):
     StatusTimeOrder = Column(Numeric(precision=12, scale=3), default=0)
     StatusSource = Column(String(32), default="Unknown")
     __table_args__ = (PrimaryKeyConstraint("JobID", "SeqNum"),)
+
+
+class TaskQueues(TaskQueueDBBase):
```
fstagni, line 215: This is just stylistic, but: you added this in the same module where JobDB (and JobLoggingDB) is. SandboxMetadataDB is instead defined in another module. We either put all "WMS-related" schemas together or we split them in different files like it was in DIRAC.

src/diracx/db/jobs/schema.py
```diff
@@ -198,3 +210,99 @@ class LoggingInfo(JobLoggingDBBase):
...
+        Integer, ForeignKey("tq_TaskQueues.TQId", ondelete="CASCADE"), primary_key=True
+    )
+    Value = Column(String(64), primary_key=True)
+    __table_args__ = (
+        Index("GridCEsTaskIndex", "TQId"),
+        Index("GridCEsValueIndex", "Value"),
+    )
+
+
+class BannedSitesQueue(TaskQueueDBBase):
```
fstagni, line 263: Would you check if this is still used anywhere?

src/diracx/db/sandbox_metadata/db.py
```diff
@@ -54,7 +56,7 @@ async def insert(
         try:
             result = await self.conn.execute(stmt)
             return result.lastrowid
-        except sqlalchemy.exc.IntegrityError:
+        except IntegrityError:
```
fstagni, line 59: Somehow I prefer the more explicit previous version.

src/diracx/db/jobs/schema.py
simon-mazenoux: This is still used by the JobScheduling optimizer (__sendToTQ method that create/fills the JobRequirements section of the JDL, which is then interpreted by the insertIntoTQ method of the JobState class which calls the TQDB insertJob method). This can be the case if a user specifies a list of banned sites and the job doesn't require any site in particular.

However, I would argue that having a BannedSitesQueue table is not very appealing. 

One solution would be to get the list of all sites, remove from this list the banned sites and insert the remaining sites in the `tq_TQToSites` table.

I think we should also ask the community if banning sites is a real use case, because this feature brings quite a lot of complexity.

fstagni: This feature is there because, most probably, of some (LHCb) user requirement of long time ago. This reminds me that we have often argued that it did not make much sense and that could even create issues. I would be tempted to remove it also.

simon-mazenoux: As long as we use the ON DELETE CASCADE constraint in DIRAC to remove the entries from this table, then I think we can remove this table from diracx.

src/diracx/db/jobs/db.py
```diff
@@ -405,3 +447,298 @@ async def get_wms_time_stamps(self, job_id):
...
+        # Oops the number of owners may have changed so we recalculate the prio for all owners in the group
+        for owner in owners:
+            await self.__set_priorities_for_entity(
+                owner,
+                group,
+                entities_shares[owner],
+                group_properties,
+                allow_background_tqs,
+            )
+            pass
```
chrisburr, line 580:
```suggestion
```

src/diracx/db/jobs/db.py
```diff
@@ -260,6 +271,37 @@ async def get_job_status(self, job_id: int) -> LimitedJobStatusReturn:
...
+        await self.conn.execute(stmt)
+
+    async def get_vo(self, job_id: int) -> str:
+        """
+        Get the VO of the owner of the job
+        """
+        # TODO: Consider having a VO column in the Jobs table
+        from DIRAC.Core.Utilities.ClassAd.ClassAdLight import ClassAd
+
+        # TODO: this is going to be problematic
```
chrisburr, line 291: Can you clarify exactly what you mean here? Just the performance or something else?

src/diracx/db/jobs/db.py
```diff
@@ -260,6 +271,37 @@ async def get_job_status(self, job_id: int) -> LimitedJobStatusReturn:
...
+            Arguments=arguments,
+            ReceptionTime=datetime.now(tz=timezone.utc),
+        )
+        await self.conn.execute(stmt)
+
+    async def get_vo(self, job_id: int) -> str:
+        """
+        Get the VO of the owner of the job
+        """
+        # TODO: Consider having a VO column in the Jobs table
```
chrisburr, line 288: This also deserves an issue in DIRAC so we can talk about it in the BiLD

src/diracx/db/jobs/db.py
```diff
@@ -405,3 +447,298 @@ async def get_wms_time_stamps(self, job_id):
...
+            )
+            #  TODO: uncomment me and understand why mypy is unhappy with join here and not elsewhere
+            # .select_from(TaskQueues.join(JobsQueue, TaskQueues.TQId == JobsQueue.TQId))
+            .where(TaskQueues.OwnerGroup == group).group_by(TaskQueues.TQId)
+        )
+        if JOB_SHARING not in properties:
+            stmt = stmt.where(TaskQueues.Owner == owner)
+        rows = await self.conn.execute(stmt)
+        # Make a dict of TQId:priority
+        tqDict: dict[int, float] = {row[0]: row[1] for row in rows}
```
chrisburr, line 611:
```suggestion
        tqDict: dict[int, float] = {tq_id: priority for tq_id, priority in rows}
```

src/diracx/db/jobs/db.py
```diff
@@ -405,3 +447,298 @@ async def get_wms_time_stamps(self, job_id):
...
+                TaskQueues.Owner,
+                TaskQueues.OwnerGroup,
+                TaskQueues.VO,
+                TaskQueues.CPUTime,
+            )
+        )
+        if tqIdList is not None:
+            stmt = stmt.where(TaskQueues.TQId.in_(tqIdList))
+
+        tqData = dict(row._mapping for row in await self.conn.execute(stmt))
```
chrisburr, line 716: I think you probably want this:
```suggestion
        tqData = dict(dict(row._mapping) for row in await self.conn.execute(stmt))
```

chrisburr, line 716: I know it's a bit painful when refactoring but we should try to move to full PEP-8 naming, e.g. `tqData` -> `tq_data`

src/diracx/db/jobs/status_utility.py
```diff
@@ -146,3 +150,195 @@ async def set_job_status(
...
+
+async def kill_job(
+    job_id: int,
+    config: Config,
+    job_db: JobDB,
+    job_logging_db: JobLoggingDB,
+    task_queue_db: TaskQueueDB,
+    background_task: BackgroundTasks,
+):
+    await __kill_delete_job(
```
chrisburr, line 198: I'd avoid double underscore method names as they provide very little value most of the time:
```suggestion
    await _kill_delete_job(
```

src/diracx/db/jobs/status_utility.py
```diff
@@ -146,3 +150,195 @@ async def set_job_status(
...
+            ),
+        ],
+        sorts=[],
+    )
+    if not res:
+        raise NoResultFound(f"Job {job_id} not found")
+
+    status = res[0]["Status"]
+    owner = res[0]["Owner"]
+    owner_group = res[0]["OwnerGroup"]
```
chrisburr, line 249: I find indexing into tuples like this kind of hard to read, how about this:
```suggestion
    job_info, = res

    status = job_info["Status"]
    owner = job_info["Owner"]
    owner_group = job_info["OwnerGroup"]
```

It also ensures we only have one result (which I know should be the case, but I'd rather be defensive and have situation like this crash instead of being tricky to debug).

src/diracx/db/jobs/status_utility.py
```diff
@@ -146,3 +150,195 @@ async def set_job_status(
...
+    vo = await job_db.get_vo(job_id)
+
+    if status == JobStatus.STAGING:
+        returnValueOrRaise(StorageManagerClient().killTasksBySourceTaskID([job_id]))
```
chrisburr, line 253: We can't use the client here. At the very least it needs to be a `raise NotImplementedError` and a `TODO`.

src/diracx/db/jobs/status_utility.py
```diff
@@ -146,3 +150,195 @@ async def set_job_status(
...
+    status = res[0]["Status"]
+    owner = res[0]["Owner"]
+    owner_group = res[0]["OwnerGroup"]
+    vo = await job_db.get_vo(job_id)
+
+    # Remove the staging task from the StorageManager
+    # TODO: this was not done in the JobManagerHandler, but it was done in the kill method
+    # I think it should be done here too
+    if status == JobStatus.STAGING:
+        returnValueOrRaise(StorageManagerClient().killTasksBySourceTaskID([job_id]))
```
chrisburr, line 316: Similarly with the need for a TODO + `NotImplementedError`

src/diracx/db/jobs/status_utility.py
```diff
@@ -146,3 +150,195 @@ async def set_job_status(
...
+    # Remove the job from SandboxMetadataDB
+    # TODO: this was also  not done in the JobManagerHandler, but it was done in the JobCleaningAgent
+    # I think it should be done here as well
+    await sandbox_metadata_db.unassign_sandbox_from_job(job_id)
+
+    # Remove the job from TaskQueueDB
+    try:
+        tq_id = await task_queue_db.get_tq_id_for_job(job_id)
+        await task_queue_db.remove_job(job_id)
+        background_task.add_task(
```
chrisburr, line 327: We should have a `TODO` here to move it to celery

src/diracx/routers/job_manager/__init__.py
```diff
@@ -406,6 +496,98 @@ async def get_single_job(job_id: int):
...
+    background_task: BackgroundTasks,
+):
+    """
+    Fully remove a job from the WMS databases.
+
+    WARNING: This endpoint has been implemented for the compatibility with the legacy DIRAC WMS
+    and the JobCleaningAgent. However, once this agent is ported to diracx, this endpoint should
+    be removed, and the delete endpoint should be used instead.
+    """
+
```
chrisburr, line 572: So it can be more easily found by grepping:
```suggestion
    # TODO: Remove once legacy DIRAC no longer needs this
```

tests/routers/test_job_manager.py
```diff
@@ -589,3 +589,60 @@ def test_set_job_status_with_invalid_job_id(normal_user_client: TestClient):
...
+
+def test_delete_job(normal_user_client: TestClient):
+    # Arrange
+    job_definitions = [TEST_JDL]
+    r = normal_user_client.post("/jobs/", json=job_definitions)
+    assert r.status_code == 200, r.json()
+    assert len(r.json()) == 1
+    job_id = r.json()[0]["JobID"]
+
+    # Act
```
chrisburr, line 602: Perhaps this would make sense as a pytest fixture to remove the repetition?

```suggestion
def test_delete_job(normal_user_client: TestClient, test_job_id: int):
```

src/diracx/routers/job_manager/__init__.py
```diff
@@ -233,14 +236,101 @@ def __init__(self, user_info: UserInfo, allInfo: bool = True):
...
+                    job_id,
+                    config,
+                    job_db,
+                    job_logging_db,
+                    task_queue_db,
+                    background_task,
+                )
+                for job_id in job_ids
+            )
+        )
```
chrisburr, line 261: Would it be practical to make this an actual bulk operation on the DB side? i.e. something like:

```
DELETE FROM JOBS WHERE JOB_ID IN (123, 124, 125, 126)
```

rather than

```
DELETE FROM JOBS WHERE JOB_ID = 123
DELETE FROM JOBS WHERE JOB_ID = 124
DELETE FROM JOBS WHERE JOB_ID = 125
DELETE FROM JOBS WHERE JOB_ID = 126
```

(This is probably better as an issue and then a separate PR rather than adding complexity to this one.)

src/diracx/routers/job_manager/__init__.py
```diff
@@ -233,14 +236,101 @@ def __init__(self, user_info: UserInfo, allInfo: bool = True):
...
+    except NoResultFound as e:
```
chrisburr, line 262: We should probably use a [`TaskGroup`](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup)+`except*` instead of `gather` so we get the better exception handling.

Ideally I think we should return a more explicit error like:

```python
detail={
    "message": "Failed to remove 5 jobs out of 167",
    "failed_ids": [1, 78, 32, 25, 90]
}
````

or maybe even:

```python
detail={
    "message": "Failed to remove 5 jobs out of 167",
    "failed_ids": {
        "1": str(e1),
        "78": str(e2),
        "32": str(e3),
        "25": str(e4),
        "90": str(e5)
    }
}
````

simon-mazenoux: Do you think we should cancel any action if one of them fail ? (for example here should we remove the other 162 jobs or not ?)

chrisburr: If that’s practical I think that would be nice. What do you think @fstagni @chaen?

src/diracx/db/jobs/status_utility.py
chrisburr: I just chatted with @chaen and he thinks we should just be a TODO and not even bother with the `NotImplementedError`.

src/diracx/routers/job_manager/__init__.py
fstagni: > If that’s practical I think that would be nice.

I concur.

chrisburr: @chaen Agrees as well

src/diracx/routers/job_manager/__init__.py
simon-mazenoux: I ended up doing it here because I had problems using the [TaskGroup](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup)+except* as you advised, and I wanted to see if making bulk methods could help me avoid theses. In the end, I still had to use them, but most of the work was already done. Sorry for the complexity.

src/diracx/routers/job_manager/__init__.py
simon-mazenoux: Linked discussion: #97

simon-mazenoux:
@chaen I think the pytest-integration tests are failing because the TaskQueueDB and SandboxMetadataDB are not in diracx-charts. Is this normal ? If so, do you want me to add them in a pull request there, or should I let you do it?

PR #67, opened 2023-09-04
Start using the integration branch of DIRAC

chrisburr:


chaen:
I think we need https://github.com/DIRACGrid/DIRAC/pull/7143 to be merged first, and then there is also another replacement to be made in #55

chaen:
Superseeded with https://github.com/DIRACGrid/diracx/pull/55

Issue #29, opened 2023-06-29
Implement JobParametersDB


chaen:
- [ ] Does this need to support both MySQL and ElasticSearch DB ?

fstagni:
Answers:
- A personal opinion, but I would not provide also the MySQL version: ElasticSearch is a pretty good choice for all Job parameters, and the pilot ones (linking here https://github.com/DIRACGrid/DIRAC/issues/6622).
- Some job attributes (e.g. (Minor)Status, few flags) are changing often until a job reaches a final status. This is not what should go in ES.
- The older index format has already been dropped in https://github.com/DIRACGrid/DIRAC/pull/7019

chrisburr:
We just decided that we should probably drop support for writing to the MySQL JobParametersDB in 8.1, will be finalised at the next BiLD.

arrabito:
In #76 I've done a very preliminary implementation of the JobParametersDB in OpenSearch.
I wait for the review, before adding new features.

fstagni:
Hi @arrabito, since the refactoring mentioned in https://github.com/DIRACGrid/diracx/pull/76#issuecomment-1717069616 is done, would you be willing to complete this task? (mind also https://github.com/DIRACGrid/DIRAC/pull/7292)

arrabito:
I will be glad to complete this task but I will be able to work on it only starting from about mid May.
Is it acceptable for you?

fstagni:
Looking at the code, and at the logs, my understanding is that the implementation is already done. Main and most recent PR is https://github.com/DIRACGrid/diracx/pull/397, logs write down "PATCH /api/jobs/metadata HTTP/1.1" 204 No Content". 

I believe this could be closed, but implementers @ryuwd and @chrisburr might remember something else that needs to be done.

PR #72, opened 2023-09-05
Feature/npigoux reschedule

natthan-pigoux:
Related issue: https://github.com/DIRACGrid/diracx/issues/22

Start creating the `rescheduleJob` method in `JobDB` as well as the routers methods and a `test_insert_and_reschedule` test. 
Multiple features are missing (commented as TODO) and the method has to be finalized when those features will be added.


src/diracx/db/jobs/db.py
```diff
@@ -1,16 +1,18 @@
...
+from sqlalchemy import delete, func, insert, select, update
 
 from diracx.core.exceptions import InvalidQueryError
+from diracx.core.models import ScalarSearchOperator, ScalarSearchSpec
 from diracx.core.utils import JobStatus
 
 from ..utils import BaseDB, apply_search_filters
 from .schema import Base as JobDBBase
-from .schema import InputData, JobJDLs, Jobs
+from .schema import InputData, JobJDLs, JobParameters, Jobs, OptimizerParameters
```
fstagni, line 15: In https://github.com/DIRACGrid/diracx/issues/29 (and at the last hackathon) we discuss against adding MySQL support for `JobParameters`. Before we ratify this decision I would, at a minimum, avoid adding this here.

fstagni: Also, I am TBH not sure that `OptimizerParameters` will be something that we'll need in DiracX. So, avoid adding code for it atm.

src/diracx/db/jobs/db.py
```diff
@@ -193,7 +214,7 @@ async def insert(
 
             return {
                 "JobID": job_id,
-                "Status": JobStatus.FAILED,
+                "Status": JobStatus.Failed,
```
fstagni, line 217: Does not seem correct to me.

src/diracx/db/jobs/db.py
```diff
@@ -141,6 +148,20 @@ async def setJobJDL(self, job_id, jdl):
         )
         await self.conn.execute(stmt)
 
+    async def getJobJDL(self, job_id: int, original: bool = False) -> str:
```
fstagni, line 151: TBH I am not sure we have decided if there will be JDLs in DiracX...

src/diracx/db/jobs/db.py
```diff
@@ -242,3 +263,149 @@ async def insert(
...
+
+        # TODO: update maxRescheduling:
+        # self.maxRescheduling = self.getCSOption("MaxRescheduling", self.maxRescheduling)
+
+        if reschedule_counter > self.maxRescheduling:
+            logging.warn(f"Job {job_id}: Maximum number of reschedulings is reached.")
+            self.setJobAttributes(
+                job_id,
+                {
+                    "Status": JobStatus.Failed,
```
fstagni, line 320:
```suggestion
                    "Status": JobStatus.FAILED,
```

?

src/diracx/db/jobs/db.py
```diff
@@ -242,3 +263,149 @@ async def insert(
...
+                    "MinorStatus": "Maximum of reschedulings reached",
```
fstagni, line 321: The cleaner way would be to add this to `JobMinorStatus` module.

fstagni (commented):
I move this to "Draft" given the To-DOs and discussions.

natthan-pigoux:
I added a `JobMinorStatus` class in models and I now insert status in the JobLoggingDB.
Waiting for JobPolicy and TaskQueueDB (https://github.com/DIRACGrid/diracx/issues/63) to continue.

src/diracx/core/models.py
```diff
@@ -67,6 +67,11 @@ class JobStatus(str, Enum):
     RESCHEDULED = "Rescheduled"
 
 
+class JobMinorStatus(str, Enum):
```
fstagni, line 70: Can be taken from https://github.com/DIRACGrid/DIRAC/blob/integration/src/DIRAC/WorkloadManagementSystem/Client/JobMinorStatus.py

chrisburr (approved):

PR #73, opened 2023-09-05
fix: add mypy option in pyproject to avoid module import errors

simon-mazenoux:
Add a mypy option in pyproject file to avoid module import errors from DIRAC

(see https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports)


chrisburr:
Closing as this was included in https://github.com/DIRACGrid/diracx/pull/55/

PR #75, opened 2023-09-06
Style updates

fstagni:


pyproject.toml
```diff
@@ -17,6 +25,12 @@ exclude = ["src/diracx/client/"]
 # Allow default arguments like, e.g., `data: List[str] = fastapi.Query(None)`.
 extend-immutable-calls = ["fastapi.Depends", "fastapi.Query", "fastapi.Path", "fastapi.Body", "fastapi.Header"]
 
+[tool.black]
+line-length = 120
```
chaen, line 29: I don't think it is a good idea here. As much as I think 120 is better than 80, the default is 80 in each and every tool. It made sense for DIRAC, but for such a new project, I'd stick to 80

fstagni: In which tools?

I am tired of having hardly-readable lines cut at 80 characters.

chrisburr: This is basically decided for us:

* https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length
* https://prettier.io/docs/en/options.html#print-width
* https://peps.python.org/pep-0008/#maximum-line-length

From my experience outside of DIRAC, if the code isn't readable with the 80-ish rule then it's often badly written and hard to read and extending the line length is putting lipstick on a pig.

Also, reading 120-charactor width code on a 13" laptop is aweful despite it being the most common form factor. Even on my 15 inch laptop I have to scroll horizontally to view 120 characters on GitHub (and that's with display scaling that most people find unreadable).

fstagni: I see this is getting opinionated. Anyway, what is "standard" today will change to a new "standard" at some point. Even `black` by default sets 88 and not 80.

Anyway, I have a 13" laptop and I prefer 120 characters. And anyway we all connect (sometimes ultra) large screens. We can just "poll" at the next BildX and see.

chrisburr: > I see this is getting opinionated.

This has started from you deciding to unilaterally push your opinion in this PR without even having a description mentioning that you changed it.

> Anyway, what is "standard" today will change to a new "standard" at some point.

And then we'll pick it up as part of an annual-black-with-breaking-changes update or whatever other formatter we switch to.

> We can just "poll" at the next BildX and see.

I might even agree that a longer length is appropriate, especially as type hints and annotations make the lines a lot longer. That said I think it would be more appropriate to wait until DiracX is a bit more fleshed out and then see what we think the best trade off is for the style of code we end up with. Until then it's just busy work in my opinion.

PR #76, opened 2023-09-07
Implement OpenSearchJobParametersDB 

arrabito:
Very preliminary implementation of OpenSearchJobParametersDB.
It allows to connect to the OpenSearch instance.
The set_single_job_parameters method allows to create an index with dummy job parameters.

### DIFF
--- a/src/diracx/db/job_parameters/db.py
+++ b/src/diracx/db/job_parameters/db.py
@@ -1,0 +1,26 @@
+from __future__ import annotations
+
+from opensearchpy import OpenSearch
+
+try:
+    from opensearchpy import A, Q, Search
+except ImportError:
+    pass
+
+from ..utils import BaseDB
+
+
+class OpenSearchJobParametersDB(BaseDB):
+    async def set_client():
+        # TODO: get connection secrets from SettingsClass.create and connection details from configuration
+        host = "opensearch-cluster-master"
+        port = 9200
+        # For now credentials are hard coded, to be set here
+        auth = ("user", "pwd")
+        client = OpenSearch(
+            hosts=[{"host": host, "port": port}],
+            http_auth=auth,
+            use_ssl=True,
+            verify_certs=False,
+        )
+        return client
--- a/setup.cfg
+++ b/setup.cfg
@@ -77,6 +77,7 @@ diracx.dbs =
 	AuthDB = diracx.db:AuthDB
 	JobDB = diracx.db:JobDB
 	SandboxMetadataDB = diracx.db:SandboxMetadataDB
+	OpenSearchJobParametersDB = diracx.db:OpenSearchJobParametersDB
 	#DummyDB = diracx.db:DummyDB
 diracx.services =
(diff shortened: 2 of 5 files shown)


arrabito:
Concerning pre-commit errors:

1.  `src/diracx/db/job_parameters/db.py:14: error: Method must have at least one argument. Did you forget the "self" argument?  [misc]`

When I add the "self" argument to:

`    async def set_client(self)`

the "self" argument is required when calling the `set_single_job_parameters` from the web interface. I don't know how to avoid this.

2. `src/diracx/routers/job_manager/__init__.py:111: error: "OpenSearchJobParametersDB" has no attribute "index"  [attr-defined`]
`src/diracx/routers/job_manager/__init__.py:113: error: "OpenSearchJobParametersDB" has no attribute "get"  [attr-defined]`

I didn't have these errors when I commited on my machine. Indeed OpenSearchJobParametersDB is adefined in dependencies.py as:

```
OpenSearchJobParametersDB = Annotated[
    _OpenSearchJobParametersDB, Depends(_OpenSearchJobParametersDB.set_client)
]
```

Also I've tested the method `set_single_job_parameters` by running a full demo on my machine and it works well.

Then just for your information, I had a pre-commit error on my machine that it was not related to my code and that I've ignored:
`src/diracx/routers/auth.py:56: error: Unexpected keyword argument "env_prefix" for "__init_subclass__" of "object"  [call-arg]`

src/diracx/db/job_parameters/db.py
```diff
@@ -0,0 +1,26 @@
...
+from ..utils import BaseDB
+
+
+class OpenSearchJobParametersDB(BaseDB):
+    async def set_client():
+        # TODO: get connection secrets from SettingsClass.create and connection details from configuration
+        host = "opensearch-cluster-master"
+        port = 9200
+        # For now credentials are hard coded, to be set here
+        auth = ("user", "pwd")
```
arrabito, line 19: Note that these are fake credentials. In order to make it work in a demo environment, one should substitute them with the correct ones.

src/diracx/db/job_parameters/db.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import annotations
+
+from opensearchpy import OpenSearch
+
+try:
+    from opensearchpy import A, Q, Search
+except ImportError:
+    pass
+
+from ..utils import BaseDB
```
fstagni, line 10: This is for sqlalchemy supported DBs, I would not mix it here.

src/diracx/db/job_parameters/db.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import annotations
+
+from opensearchpy import OpenSearch
```
fstagni, line 3: `opensearch-py` should be added to environment.yml file

fstagni (commented):
One of the first thing to do is to have a test.

arrabito:
Thanks @fstagni and @chrisburr for your feedback.
So I wait for the refactoring before working again on this.

PR #83, opened 2023-09-12
Fix JobDB schema

chrisburr:


simon-mazenoux:
Well spotted, in DIRAC those are separate indexes, not a unique composite index.

simon-mazenoux (approved):

fstagni (approved):

PR #104, opened 2023-09-25
Add CLI tests

chrisburr:


aldbr:
> ddev: closing it because very old and probably easier to restart from scratch + CLI is still experimental for now. Let's wait for a real usage/use case

PR #106, opened 2023-09-26
Add userinfo endpoint

chrisburr:
Also includes a crude `dirac whoami` command.

### DIFF
--- a/src/diracx/routers/auth.py
+++ b/src/diracx/routers/auth.py
@@ -994,3 +994,24 @@ async def authorization_flow_complete(
     return responses.RedirectResponse(
         f"{redirect_uri}?code={code}&state={decrypted_state['external_state']}"
     )
+
+
+class UserInfoResponse(TypedDict):
+    sub: str
+    vo: str
+    dirac_group: str
+    properties: list[SecurityProperty]
+    preferred_username: str
+
+
+@router.get("/userinfo")
+async def userinfo(
+    user_info: Annotated[UserInfo, Depends(verify_dirac_access_token)]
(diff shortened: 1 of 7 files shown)


src/diracx/routers/auth.py
```diff
@@ -994,3 +994,24 @@ async def authorization_flow_complete(
...
+    dirac_group: str
+    properties: list[SecurityProperty]
+    preferred_username: str
+
+
+@router.get("/userinfo")
+async def userinfo(
+    user_info: Annotated[UserInfo, Depends(verify_dirac_access_token)]
+) -> UserInfoResponse:
+    return {
```
aldbr, line 1011: Out of curiosity: `UserInfoResponse` is similar to `UserInfo` in terms of content.
Would it make sense to have such inheritance?

```
UserInfo <- AuthInfo (could be renamed AuthUserInfo)
         <- UserInfoResponse
```

chrisburr: I hadn't considered that, my first instinct was yes but then I realised the `properties` field in the `AuthInfo` class so we would need to duplicate it. I now lean towards treating the return type as an independent thing but I'm open to other opinions.

aldbr: Yes indeed, you're right.
An alternative would be:

```
UserInfo <- UserInfoResponse (UserPropertyInfo?) <- AuthInfo (AuthUserInfo)
```

But then it makes a lot of inheritance for just a few attributes and we probably lose clarity.

chaen (approved):

PR #110, opened 2023-09-27
Add support for uploading/downloading sandboxes

chrisburr:
```python
from pathlib import Path
from diracx.client.aio import DiracClient
import diracx.api

async with DiracClient() as client:
    pfn = await diracx.api.jobs.create_sandbox(client, [Path("/tmp/recompress-part1.txt"), Path("/tmp/recompress-meta-part2.zip")])
    print(pfn)

destination = Path("/tmp/test")
async with DiracClient() as client:
    await diracx.api.jobs.download_sandbox(client, pfn, destination)
assert (destination / "recompress-part1.txt").is_file()
assert (destination / "recompress-meta-part2.zip").is_file()
```


src/diracx/api/jobs.py
```diff
@@ -0,0 +1,83 @@
...
+            checksum=checksum,
+            size=os.stat(tar_fh.fileno()).st_size,
+            format=f"tar.{SANDBOX_COMPRESSION}",
+        )
+
+        res = await client.jobs.initiate_sandbox_upload(sandbox_info)
+        if res.url:
+            logger.debug("Uploading sandbox for %s", res.pfn)
+            files = {"file": ("file", tar_fh)}
+            response = httpx.post(res.url, data=res.fields, files=files)
```
chaen, line 55: await

PR #116, opened 2023-10-02
Update SandboxMetadataDB interface

chrisburr:
Ready for review but needs #114. Only the last commit is unique to this PR.


src/diracx/core/models.py
```diff
@@ -112,3 +112,18 @@ class UserInfo(BaseModel):
...
+    SHA256 = "sha256"
+
+
+class SandboxFormat(StrEnum):
+    TAR_BZ2 = "tar.bz2"
+
+
+class SandboxInfo(BaseModel):
+    checksum_algorithm: ChecksumAlgorithm
+    checksum: str = Field(pattern=r"^[0-f]{64}$")
```
chaen, line 127: This pattern only holds true for sha256, so it is sort of defeats the purpose doesn't it ? 
Either we can set the checksum algorithm to `Literal["sha256"]`  or we should have a class mapping the checksum algorithm with its pattern

chrisburr: It would also work for any other 256-bit hash like `sha3_256`/`shake_256`/`blake2s` so it's likely even if we switched alogrithm we'd still keep the same pattern as we wouldn't need any more entropy. Regardless it's an easy refactoring to make it more generic if it was ever needed.

chaen (approved):

PR #126, opened 2023-10-03
Fixes for getting DiracX installed on lbcertifidirac70

chrisburr:
See the commit messages for details.

### DIFF
--- a/src/diracx/routers/auth.py
+++ b/src/diracx/routers/auth.py
@@ -481,7 +481,7 @@ async def do_device_flow(
     (note: it can't be put as parameter or in the URL)
     """
 
-    # Here we make sure the user_code actualy exists
+    # Here we make sure the user_code actually exists
     scope = await auth_db.device_flow_validate_user_code(
         user_code, settings.device_flow_expiration_seconds
     )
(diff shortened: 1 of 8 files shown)


src/diracx/routers/auth.py
```diff
@@ -348,6 +354,7 @@ async def initiate_device_flow(
             status_code=status.HTTP_400_BAD_REQUEST, detail="Unrecognised client ID"
         )
 
+    # TODO: This is an information leak
```
chaen, line 357: what is ? The exception we return ?

src/diracx/routers/job_manager/sandboxes.py
```diff
@@ -144,7 +152,7 @@ def pfn_to_key(pfn: str) -> str:
 
 SANDBOX_PFN_REGEX = (
     # Starts with /S3/<bucket_name>
```
chaen, line 154: this comment needs updating

src/diracx/routers/auth.py
chrisburr: Ignore me, I didn't think about it properly 😄

src/diracx/routers/job_manager/sandboxes.py
chrisburr: Fixed

chaen (approved):

PR #127, opened 2023-10-04
feat: include components to communicate with the webapp

chaen:
Introduce the necessary components to include `diracx-web` in `diracx-charts`.

### DIFF
--- a/src/diracx/routers/__init__.py
+++ b/src/diracx/routers/__init__.py
@@ -9,6 +9,7 @@ from typing import Any, AsyncContextManager, AsyncGenerator, Iterable, TypeVar
 import dotenv
 from fastapi import APIRouter, Depends, Request
 from fastapi.dependencies.models import Dependant
+from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, Response
 from fastapi.routing import APIRoute
 from pydantic import parse_raw_as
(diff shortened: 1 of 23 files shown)


src/diracx/routers/__init__.py
```diff
@@ -143,6 +143,20 @@ def create_app_inner(
     app.add_exception_handler(DiracError, dirac_error_handler)
     app.add_exception_handler(DiracHttpResponse, http_response_handler)
 
+    from fastapi.middleware.cors import CORSMiddleware
+
+    origins = [
+        "http://localhost:8000",
```
aldbr, line 149: Should we create an environment variable to provide the `CORSMiddleware` with the `origins`?

As this is a temporary solution (:crossed_fingers: ), I prefer asking.

chrisburr: It's also possible to inject this in the ingress controller so we might be best of letting it be an option of the helm chart rather than DiracX itself. (Though the hack is good enough for now.)

chrisburr: (i.e. nginx would include extra headers in the response)

aldbr: Yes you're right, I will try to set this up in the ingress controller, thanks!

src/diracx/routers/fastapi_classes.py
```diff
@@ -59,6 +61,8 @@ def __init__(
         *,
         dependencies=None,
         require_auth: bool = True,
+        path_root: str = "/api",
```
aldbr, line 64: Is it still needed?

src/diracx/routers/__init__.py
aldbr: Is the `CORSMiddleware` still needed now?

src/diracx/db/sql/jobs/schema.py
```diff
@@ -92,6 +92,7 @@ class Jobs(JobDBBase):
     AccountedFlag = Column(
         "AccountedFlag", Enum("True", "False", "Failed"), default="False"
     )
+    VO = Column("VO", String(64), default="Unknown")
```
aldbr, line 95: I think this is already done in this PR, which is ready to be merged by the way: https://github.com/DIRACGrid/diracx/pull/102/files

src/diracx/routers/__init__.py
chrisburr: Not for now at least.

chrisburr (approved):

PR #151, opened 2023-10-18
Minor pre-commit fixes.

fstagni:


.pre-commit-config.yaml
```diff
@@ -17,18 +17,20 @@ repos:
     hooks:
       - id: ruff
         args: ["--fix"]
+        exclude: ^(src/diracx/client/)
 
   - repo: https://github.com/psf/black
     rev: 23.9.1
     hooks:
       - id: black
+        exclude: ^(src/diracx/client/)
```
chrisburr, line 26: Why? The client generation isn't being modifed and relies on this to format it. Also there is manually written code in `client/` (mainly the patch files) which should definitely be linted and formatted.

src/diracx/cli/__init__.py
```diff
@@ -22,10 +21,9 @@
 @app.async_command()
 async def login(
     vo: str,
-    group: Optional[str] = None,
-    property: Optional[list[str]] = Option(
-        None, help="Override the default(s) with one or more properties"
-    ),
+    group: str | None = None,
```
chrisburr, line 24: Is the typer problem fixed?

.pre-commit-config.yaml
fstagni: From previous discussions the suggestion was to avoid linting this dir. Anyway, fine for me to remove this line.

src/diracx/cli/__init__.py
fstagni: Unfortunately not, so I removed the forced changes.

chrisburr (approved):

PR #155, opened 2023-10-20
Add distinct option to job search

### LINKED ISSUE #134: Add a distinct option to the db search method
The use case is to be able to return the distinct list of values matching given parameter. This fits nicely in the `search` method of the DB classes, that needs adding a `distinct=False` parameter.

* [ ] Add the parameter to the route (see   https://github.com/DIRACGrid/diracx/blob/39a73bbd84e8baf96a58d8b7058c2bb0fc1966be/src/diracx/routers/job_manager/__init__.py#L52 )
* [ ] Add the behavior 
* [ ] Add the tests

sfayer:
This adds a distinct option to the job search interface (and db class) plus a test; I hope that's what you had in mind for #134?

Closes #134 (maybe).

### DIFF
--- a/src/diracx/db/sql/jobs/db.py
+++ b/src/diracx/db/sql/jobs/db.py
@@ -68,7 +68,14 @@ class JobDB(BaseSQLDB):
         ]
 
     async def search(
-        self, parameters, search, sorts, *, per_page: int = 100, page: int | None = None
+        self,
+        parameters,
+        search,
+        sorts,
+        *,
+        distinct: bool = False,
+        per_page: int = 100,
+        page: int | None = None,
     ) -> list[dict[str, Any]]:
         # Find which columns to select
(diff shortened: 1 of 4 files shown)


src/diracx/db/sql/jobs/db.py
```diff
@@ -68,7 +68,14 @@ async def summary(self, group_by, search) -> list[dict[str, str | int]]:
         ]
 
     async def search(
-        self, parameters, search, sorts, *, per_page: int = 100, page: int | None = None
+        self,
+        parameters,
+        search,
+        sorts,
+        distinct=False,
+        *,
```
chaen, line 76:
```suggestion
        *,
        distinct: bool = False,
```

chaen:
Perfect thanks ! I'll just push the client regeneration to this branch, and then merge it :-)

PR #158, opened 2023-10-24
tmp handler except

chaen:
Need to resuffle the initialization to be able to reconnect


environment.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
   - aiohttp
   - aiomysql
   - aiosqlite
+  - asyncache
```
chrisburr, line 11: I'm not sure there is enough code in this package to justify it as a dependency. It also doesn't seem to be that widely used or actively maintained.

PR #164, opened 2023-10-26
Use diracx/server-base:latest as a basis of the docker image

chrisburr:
Part of #163


chrisburr:
I'm expecting this to break the helm chart but I'm inclinced to suggest we just merge and I'll then fix it.

chaen (approved):

PR #173, opened 2023-11-04
Split into multiple namespace packages

### LINKED ISSUE #170: Split installation into diracx, diracx-server, ...

chrisburr:
Closes #170


chrisburr:
What do you mean by "different packages"? Do you mean like different repositories? Or different top level imports? Or something else?

fstagni:
Sorry. Different repositories.

chrisburr:
Having to open multiple PRs across different repositories can be quite painful. It then becomes hard to keep them in sync for testing and you have to tag them each individually.

I think this is the way that we can get the benifit of having partial installations while minimally affecting developers.

I'm open to discussing more if you have thoughts.

fstagni:
I don't know what's the best way, I am just trying to have a more clear idea.

Supposing we split in different repositories, how many of them would there be? IUUC here you are splitting into:
- diracx-api
- diracx-cli
- diracx-client
- diracx-core
- diracx-db
- diracx-routers
- diracx-testing

Naively, I have the impression that *-db* and *-routers* could be joined into *-server* (and *-testing* too?) and that *-api*, *-cli* could go into *-client*.

The fact of tagging each one separately might not be a disadvantage.

chaen:
We had long discussions about that (even before we presented DiracX for the first time). We ended up with the conclusion that splitting things as the are offers us the best of two worlds:
* it's easy to make sure the dependencies don't escape (e.g. you don't need SQLAlchemy in the client). 
* We can have specific installations (e.g. only for tasks, only for routers, only for client)
* Having everything in the same repo makes it infinitely easier to manage. We deemed way too complicated managing multiple tags (for compatibility management, and risk of mistakes).

What advantage would you see in having different repo/tags ? We could not see any, but maybe you do ?

chrisburr:
When thinking about the layout you should also consider `diracx-tasks` which will potentially depend on `diracx-db` and/or `diracx-client`+`diracx-api` (we haven't decided which route we would want to take and we might choose both). This way:
* `diracx-tasks` depends on `celery`+`gfal2` but not `fastapi`
* `diracx-routers` depends on `fastapi` but not `celery`+`gfal2`

fstagni:
(for the record: this will be discussed at next BildX.)

PR #185, opened 2023-11-26
Fix legacy integration tests

chrisburr:


codecov.yml
```diff
@@ -12,6 +12,6 @@ coverage:
 
 codecov:
   notify:
-    after_n_builds: 2
+    after_n_builds: 7
```
chaen, line 15: why 7 ?

containers/client/Dockerfile
```diff
@@ -1,11 +1,8 @@
 FROM ghcr.io/diracgrid/diracx/client-base
 
-# Needs to specify how we install diracx
-# * from a local wheel package: PACKAGES_TO_INSTALL=/path/to/diracx.whl
-# * from pypi "diracx==v0.a.b"
-ARG PACKAGES_TO_INSTALL
+ARG EXTRA_PACKAGES_TO_INSTALL
```
chaen, line 3: we should have a default here

codecov.yml
chrisburr: 6 unit tests + the one with the demo

containers/client/Dockerfile
chrisburr: Just `DIRAC` then?

chaen: and all the `diracx` modules (not only for this DOckerifle)

chaen (approved):

PR #186, opened 2023-11-27
feat: internals `add_vo()` and `add_group()`

aldbr:
Would this make sense?

It would allow setting up multiple VOs and groups when launching `diracx` with `run_demo` (of course we would need another PR in `diracx-charts`).
- The alternative is to manually change the config (`default.yml`) and commit it from a running pod (which is probably fine too).


chaen:
Yes it makes sense, but we are just about to break and fix the tests, the packaging, and helm, so if it can wait for tomorrow to be merged, it would be great :-)

PR #187, opened 2023-11-27
feat: disable refresh token rotation if token is for legacy DIRAC

aldbr:


chrisburr (approved):

chrisburr:
That's a much nicer solution than what I had in mind!

Issue #207, opened 2024-01-28
Ensure gConfig is never initialised


chrisburr:
We can now set `DIRAC_DISABLE_GCONFIG_REFRESH` to ensure that DIRAC crashes instead of letting `gConfig` be accessed (which is always an issue as it's not initialised when using DiracX).

- [ ] Disable for the unit tests
- [ ] Disable in the container images

PR #208, opened 2024-01-28
Ensure legacy DIRAC CS is never initialised

chrisburr:
Part of https://github.com/DIRACGrid/diracx/issues/207

Leaving as draft as it needs DIRAC to be fixed first https://github.com/DIRACGrid/DIRAC/issues/7426


chrisburr:
No longer relavent now that we have DIRACCommon

Issue #30, opened 2023-06-29
Authorization: still to do


chaen:
- [ ] Figure out the `audience` properly (especially in init of the authorization code flow)
- [ ] Encrypt the state
- [ ] Hash some fields in the db (auth code flow and device code)
- [x] Use asymeteric crypto for JWT we mint
- [ ] validate poperties against subject (in `exchange_token`)
- [ ] document the flows used
- [ ] add tests against IAM
- [ ] add tests for internal validation (tokens, scopes, etc, expiry, signing)
- [ ] One ClientID per VO ? (i.e. one swagger for all) or one clientID for all (i.e. one swagger per vo)
- [ ] One client per group ? vo in the scope instead of URL ?
- [ ] What happens if a user is suspended in IAM ( does that exists?)

aldbr:
> Figure out the audience properly (especially in init of the authorization code flow)

I see the rationale behind including the `audience` claim within the DIRAC token payload, in line with the JWT standard detailed in [RFC 7519 Section 4.1.3](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3). However, the necessity of specifying an `audience` parameter at the initial stages of both the device flow and the authorization code flow puzzles me, especially since neither [RFC 6749 Section 4.1.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1) nor [RFC 8628 Section 3.1](https://datatracker.ietf.org/doc/html/rfc8628/#section-3.1) lists `audience` as a requisite parameter.

While IAM solutions like Indigo IAM or Auth0 incorporate it to distribute tokens for various external services, our scenario appears to be different, given our current use cases and architecture.
Do we foresee any scenarios where external services would require DIRAC access tokens, justifying the inclusion of the `audience` parameter right from the start?

If not, a practical approach could be to define the `audience` claim as the DIRAC server URL (e.g. as specified by `DIRACX_TOKEN_ISSUER`) within our token payloads.

aldbr:
> What happens if a user is suspended in IAM ( does that exists?)

Currently, a same user can belong to different VOs, but this will be different in diracX: a physical user belonging to multiple VOs will be represented as different users within the middleware.

In Dirac, because a user belongs to different VOs, he/she can be suspended by different VOs at the same time:
- userA belongs to VO1, VO2, VO3: VO1 and VO3 suspend his/her account
- userA will have a `Suspended = [VO1, VO3]` field attached to him/her

[...]

PR #215, opened 2024-02-01
Hash & Encrypt: auth module

aldbr:
Aims at solving 2 points from https://github.com/DIRACGrid/diracx/issues/30

Note to myself: Once the PR is approved, I would need to create a PR in `diracx-charts` to add the `state_key` as a Kubernetes secret.

### DIFF
--- a/diracx-core/src/diracx/core/settings.py
+++ b/diracx-core/src/diracx/core/settings.py
@@ -16,7 +16,7 @@ from pydantic import AnyUrl, BaseSettings, SecretStr, parse_obj_as
 if TYPE_CHECKING:
     from pydantic.config import BaseConfig
     from pydantic.fields import ModelField
-
+from cryptography.fernet import Fernet
 
 T = TypeVar("T")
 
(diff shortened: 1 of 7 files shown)


diracx-core/src/diracx/core/settings.py
```diff
@@ -16,7 +16,7 @@
 if TYPE_CHECKING:
     from pydantic.config import BaseConfig
     from pydantic.fields import ModelField
-
+from cryptography.fernet import Fernet
```
chrisburr, line 19: Purely for my curiosity as I haven't come across fernet before, is there a particular reason for having picked it?

aldbr: No particular reason, I just googled "python encryption" and found many examples of `Fernet`. It sounds very easy to use and well maintained.
I also noticed we were already using `cryptography` within the code.

Would you recommend any another library?

chrisburr: We should definitely use something from `cryptography`. I would have defaulted to `AES-GCM-256` but it looks like [their docs recommend `Fernet`](https://cryptography.io/en/latest/) so that's good enough for me!

chrisburr (approved):

PR #225, opened 2024-03-14
test: bandit CI

chaen:
Add the Bandit check in `ruff`.
These are basically best practices to avoid obvious security loophole. It comes with quite some false positive, but it is better than false negative when it comes to security :-)

### DIFF
--- a/diracx-routers/src/diracx/routers/auth/utils.py
+++ b/diracx-routers/src/diracx/routers/auth/utils.py
@@ -73,7 +73,7 @@ class GrantType(StrEnum):
 
     authorization_code = "authorization_code"
     device_code = "urn:ietf:params:oauth:grant-type:device_code"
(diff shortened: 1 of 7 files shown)


diracx-routers/src/diracx/routers/auth/utils.py
```diff
@@ -73,7 +73,7 @@ class GrantType(StrEnum):
 
     authorization_code = "authorization_code"
     device_code = "urn:ietf:params:oauth:grant-type:device_code"
-    refresh_token = "refresh_token"
+    refresh_token = "refresh_token"  # noqa: S105   # False positif of Bandit about hard coded password
```
chrisburr, line 76: either fix the typo or remove the comment

```suggestion
    refresh_token = "refresh_token"  # noqa: S105
```

chrisburr (approved):

PR #241, opened 2024-05-16
fix(client): better error when refresh token is expired

aldbr:
Aims at fixing the following error:

```bash
$ dirac login diracAdmin
# do some stuff, then stop using it for a few days, the refresh token expired
$ dirac jobs submit job.jdl
RuntimeError: An issue occured while refreshing your access token: Invalid JWT: bad_signature:
$ dirac login diracAdmin
RuntimeError: An issue occured while refreshing your access token: Invalid JWT: bad_signature:
$ dirac logout
$ dirac login diracAdmin
# Login successful!
```

In this PR, we check the validity of the refresh token before refreshing the access token. We should have:

```bash
$ dirac login diracAdmin
# do some stuff, then stop using it for a few days, the refresh token expired
$ dirac jobs submit job.jdl
HttpResponseError: Operation returned an invalid status 'Forbidden'
Content: {"detail":"Not authenticated"}
$ dirac login diracAdmin
# Login successful!
```

### DIFF
--- a/diracx-client/src/diracx/client/_patch.py
+++ b/diracx-client/src/diracx/client/_patch.py
@@ -8,8 +8,9 @@ Follow our quickstart for examples: https://aka.ms/azsdk/python/dpcodegen/python
 """
 from __future__ import annotations
 
-from datetime import datetime
+from datetime import datetime, timezone
 import json
+import jwt
 import requests
 
 from pathlib import Path
@@ -113,9 +114,12 @@ class DiracBearerTokenCredentialPolicy(BearerTokenCredentialPolicy):
 
         if not self._token:
             credentials = json.loads(self._credential.location.read_text())
-            self._token = self._credential.get_token(
-                "", refresh_token=credentials["refresh_token"]
-            )
+            refresh_token = credentials["refresh_token"]
+            if not is_refresh_token_valid(refresh_token):
+                # If we are here, it means the refresh token is not valid anymore
+                # we suppose it is not needed to perform the request
+                return
+            self._token = self._credential.get_token("", refresh_token=refresh_token)
 
         request.http_request.headers["Authorization"] = f"Bearer {self._token.token}"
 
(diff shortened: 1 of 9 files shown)


diracx-client/src/diracx/client/_patch.py
```diff
@@ -242,6 +246,17 @@ def get_token(location: Path, token: AccessToken | None) -> AccessToken | None:
...
 
+def is_refresh_token_valid(refresh_token: str) -> bool:
+    """Check if the refresh token is still valid."""
+    # Decode the refresh token
+    refresh_payload = jwt.decode(refresh_token, options={"verify_signature": False})
+    if not refresh_payload or "exp" not in refresh_payload:
+        return False
+
+    # Check the expiration time
+    return refresh_payload["exp"] > datetime.utcnow().timestamp()
```
chaen, line 257: `utcnow` is deprecated https://docs.python.org/3.11/library/datetime.html#datetime.datetime.utcnow

diracx-client/src/diracx/client/_patch.py
```diff
@@ -113,9 +114,12 @@ def on_request(
 
         if not self._token:
             credentials = json.loads(self._credential.location.read_text())
-            self._token = self._credential.get_token(
-                "", refresh_token=credentials["refresh_token"]
-            )
+            refresh_token = credentials["refresh_token"]
+            if not is_refresh_token_valid(refresh_token):
```
chaen, line 118: would it make sense to print a clear error message, to invite people to login again ?

aldbr: There is the possibility to add the following method in `DiracBearerTokenCredentialPolicy`:

```python
async def on_response(self, request: PipelineRequest, response: PipelineResponse) -> Awaitable[None] | None:
        if response.http_response.status_code == 401:
            print(f"Not authenticated, please log in again: {response.http_response.text()}")
            return

        return super().on_response(request, response)
```

It is executed after each request performed by `DiracClient`. However, I wonder if it makes sense to add the `print` statement in `diracx-client` as it can theoretically be used in various contexts (`diracx-cli`, `diracx-api`).

May be adding a decorator to handle authN/authZ exceptions (401, 403) and print a clear message for the end users of `dirac-cli` could be considered?

chrisburr: I think this could be included here:

https://github.com/DIRACGrid/diracx/blob/608487d09dd74dd0003e6518019d6a4dc4303e39/diracx-cli/src/diracx/cli/utils.py#L11

aldbr: https://github.com/DIRACGrid/diracx/pull/241/files#diff-f82a792c5910fa04a412a937ead18e2eb741c28fde1e6c7485b3df198b970ad7R18-R24

chrisburr (approved):

PR #243, opened 2024-05-23
feat(jobs): pagination

aldbr:
Here I provide a first implementation of the pagination mechanism for the jobs, mostly based on https://github.com/DIRACGrid/diracx/pull/6.

As I explored different examples, it became clear to me that there is no one-size-fits-all solution for pagination. Instead, various implementation possibilities exist, each offering unique approaches at different stages. 

#### Pagination Strategies:

I considered two primary pagination strategies:

- Cursor-based pagination relies on opaque tokens (cursors) to navigate through results. Clients provide a cursor to fetch the next set of results, making it suitable for real-time data and ensuring stable pagination despite changes in the dataset. Such strategy requires additional logic to manage cursors and handle edge cases and is not suitable to easily jump to specific pages.

- Page-based pagination divides results into fixed-size pages, allowing clients to request specific pages using page numbers. This approach is simpler to implement but may suffer from inconsistencies if the dataset changes frequently. It is less efficient for large datasets.

My opinion: we should prioritize simplicity for the time being. Our primary requirement is to retrieve the last 10, 100, or 1,000 jobs. Occasionally, it's useful to jump to a different page, such as when checking if a particular issue has occurred before. While there may be some cases where we need to fetch a large number of jobs at once, such instances are rare. Therefore, minor inconsistencies should not pose a significant problem. I would rather choose the page-based pagination.

#### Metadata Conveyance:

Conveying metadata to clients is essential for effective navigation as we return partial results. Common methods include Web Linking, Content-Range headers, and embedding metadata within the JSON response.

- Web Linking (used by github): https://datatracker.ietf.org/doc/html/rfc8288. This method provides links to the first, previous, current, next, and last pages in the `Link` header. While convenient for navigation, it requires clients to parse HTTP headers.

```bash
Link: <https://dirac/api/jobs?page=2&per_page=100>; rel="prev", <https://dirac/api/jobs?page=4&per_page=100>; rel="next", <https://dirac/api/jobs?page=515&per_page=100>; rel="last", <https://dirac/api/jobs?page=1&per_page=100>; rel="first"
```

- Content-Range Header: https://www.rfc-editor.org/rfc/rfc9110.html#section-14.4. This approach is similar to web linking but much more compact, where the first and last items, along with the total number of items, are provided through the Content-Range header.

```bash
Content-Range: <unit> <first item>-<last item>/<total>
Content-Range: jobs 1-10/100
```

- Envelope in the JSON: https://jsonapi.org/format/#fetching-pagination. This method involves directly including metadata along with the data in the response. While this ensures that all clients have easy access to pagination details, it also increases the complexity of the payload.

My opinion:  I initially implemented the Content-Range approach based on https://github.com/DIRACGrid/diracx/pull/6. However, I believe that web linking could also be beneficial as it would fit perfectly with the pagination parameters. Including metadata directly in the JSON is straightforward, but it would require additional parsing.


aldbr:
After trying to link the pages to `diracx-web`, I realized that page-based pagination combined with the `Content-Range` header was enough.

fstagni:
> I would rather choose the page-based pagination.

On a first look, I agree.

chrisburr:
I think we should use page-based pagination for most things but have the option of having a reliable pagination for critical operations. Perhaps we could make it simplier by having the option of:

* passing `after` instead of `page`
* `after` is only supported if sorting by the primary key of the table to keep the implementation simple
* the recommended way of iterating programatically would be with Web Linking so the use of `after` becomes an implementation detail

For inspiration:

* https://developers.facebook.com/docs/graph-api/results/
* https://docs.stripe.com/api/pagination

aldbr:
> I think we should use page-based pagination for most things but have the option of having a reliable pagination for critical operations. Perhaps we could make it simplier by having the option of:
> - passing after instead of page
> - after is only supported if sorting by the primary key of the table to keep the implementation simple

@chrisburr what you are describing is a (simple) cursor-based pagination if I understand correctly.
Do you suggest we should implement both methods (at least for the critical operations)?

> the recommended way of iterating programatically would be with Web Linking so the use of after becomes an implementation detail

Indeed, if we go with cursor-based pagination, this would be a nice way of iterating through the previous and next pages without including the cursor within the json result.

chaen:
@chrisburr and I had a chat and I think we are on the same page: go for the page based approach

aldbr:
Alright, then you can start reviewing the PR

diracx-cli/tests/test_jobs.py
```diff
@@ -1,13 +1,128 @@
...
+    with tempfile.NamedTemporaryFile(
+        mode="w", encoding="utf-8", delete=False
+    ) as temp_file:
+        temp_file.write(TEST_JDL)
+        temp_file.flush()
+        jdl_filename = temp_file.name
+
+    yield jdl_filename
+
+    os.remove(jdl_filename)
```
chrisburr, line 45: I think this can be simplified to one of these:

```suggestion
@pytest.fixture
async def jdl_file():
    with tempfile.NamedTemporaryFile(mode="w", encoding="utf-8") as temp_file:
        temp_file.write(TEST_JDL)
        temp_file.flush()
        yield temp_file.name
```

else, if using `delete=False` the flush is pointless:

```suggestion
@pytest.fixture
async def jdl_file():
    with tempfile.NamedTemporaryFile(
        mode="w", encoding="utf-8", delete=False
    ) as temp_file:
        temp_file.write(TEST_JDL)

    yield temp_file.name

    os.remove(temp_file.name)
```

chrisburr (approved):

PR #251, opened 2024-06-05
Suppport Pydantic 2

chrisburr:


chrisburr:
There is a chicken-and-egg problem with this PR between building the dev container with https://github.com/DIRACGrid/container-images/pull/16 and running the CI for this.

PR #254, opened 2024-06-09
Support mocking JobParametersDB for unit tests

chrisburr:


diracx-db/src/diracx/db/sql/utils.py
```diff
@@ -55,13 +57,61 @@ def sqlite_utcnow(element, compiler, **kw) -> str:
     return "DATETIME('now')"
 
 
+class date_trunc(expression.FunctionElement):
```
chaen, line 60: you really need to explain what you are doing here :-D

diracx-db/src/diracx/db/sql/utils.py
```diff
@@ -212,10 +264,62 @@ async def ping(self):
...
+                r"^(\d{4})-?(\d{2})?-?(\d{2})?[ T]?(\d{2})?:?(\d{2})?:?(\d{2})?\.?(\d{6})?Z?$",
+                pattern,
+                value,
+            ),
+        )
+
+    raise InvalidQueryError(f"Cannot parse {value=}")
+
+
+def apply_search_filters(column_mapping, stmt, search):
```
chaen, line 297: as a general consideration, it's probably better to give the `table` instead of the column mapping. The reason is that if SQLAlchemy changes the internals (and it certainly wouldn't be the first time), we would only need to fix it here, as opposed to everywhere we call this method. It's also more readable

diracx-db/src/diracx/db/sql/utils.py
```diff
@@ -212,10 +264,62 @@ async def ping(self):
...
     for query in search:
-        column = table.columns[query["parameter"]]
+        try:
+            column = column_mapping(query["parameter"])
+        except KeyError as e:
+            raise InvalidQueryError(f"Unknown column {query['parameter']}") from e
+
+        if isinstance(column.type, DateTime):
+            if "value" in query and isinstance(query["value"], str):
```
chaen, line 305: What would be a search case without `value` ? :thinking:

diracx-testing/src/diracx/testing/routers.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+from contextlib import asynccontextmanager
+from functools import partial
+
+
+def fake_available_implementations(name, *, real_available_implementations):
```
chaen, line 7: why do you need to redefine it here as well ?

diracx-db/src/diracx/db/sql/utils.py
chrisburr: This is something I plan to refactor to make it so we can handle joins (which we already need) so I propose to do it in my next PR?

diracx-db/src/diracx/db/sql/utils.py
chrisburr: `parameter in values` (plural!)

diracx-db/src/diracx/db/sql/utils.py
chaen: Works with me !

diracx-db/src/diracx/db/sql/utils.py
chaen: Ouh... that is actually super error prone no ? Should we use a single word for both  case, and even a constant for it ? something like `SEARCH_VALUE="value"` ?

chrisburr: Well `"value"` is always a scalar and `"values"` is always a list of scalars.

The schema of the search body can be validated to make sure you only use the right key with the right kind of operator.

fstagni:
There are conflicts to be resolved.

chaen (approved):

PR #265, opened 2024-06-24
feat(cli): update helm values in internal legacy CLI

aldbr:
Needs https://github.com/DIRACGrid/diracx-charts/pull/105


fstagni (approved):
Given that https://github.com/DIRACGrid/diracx-charts/pull/105 is merged since long time, and to my (still limited) knowledge it seems OK, I approve and merge.

Issue #258, opened 2024-06-13
Make the client send its version when calling the server


chaen:
The server should given an error if the client version is too old

The minimum version should be configured in the pyproject as a metadata such that the extension can overwrite it (`diracx.min_client_version`)

chrisburr:
* `diracx-client/src/diracx/client/_patch.py`
* `diracx-client/src/diracx/client/aio/_patch.py`
* https://fastapi.tiangolo.com/tutorial/middleware/?h=middle
* `diracx-routers/pyproject.toml`

```toml
[project.entry-points."diracx.min_client_version"]
diracx = 1.2.3
lhcbdirac = 
```

chaen:
Side note: I don't think we will have millions of them, but do we want the diracx headers to be easily distinguishable from others ? That is, do we want e.g. `DiracX-Client-Version` more than `client-version` ? 
It's nice when it's easy to spot so I would be tempted to add an extra prefix. The  `X-` standard is *deprecated* though (https://www.rfc-editor.org/rfc/rfc6648#page-4)

Also, this raises the question: what do we do in case there isn't a version ? (swagger, diracx-web)

natthan-pigoux:
Yes, I already changed the name with `X-Diracx-Client-Version`. I will remove the X- then.
I also added a TODO with the question on what to do in that cas. 
Thanks again for your help @chaen :)

aldbr:
> what do we do in case there isn't a version ? (swagger, diracx-web)

I guess the server should check the `DiracX-Client-Version` only if it is provided by the client, shouldn't it?

natthan-pigoux:
Another point is that the entrypoints in the toml for the "min_client_version" cannot be a static string but:
[project.entry-points."diracx.min_client_version"]
diracx = "diracx.routers:DIRACX_MIN_CLIENT_VERSION"

I hard coded it there for now: https://github.com/DIRACGrid/diracx/pull/267/commits/d0ff92f1cabae2c1ceb9e1101be2fee5109faba6#diff-424d2a991980de5065c3b1e58f5066bf78641a4804b34bb5977be2ad95289ae7R483

PR #267, opened 2024-06-27
feat: send client version to server

natthan-pigoux:
- create a middleware to check that the client version pass in header is recent enough from what is defined in the entry_point

### DIFF
--- a/diracx-routers/pyproject.toml
+++ b/diracx-routers/pyproject.toml
@@ -56,6 +56,8 @@ auth = "diracx.routers.auth:router"
 WMSAccessPolicy = "diracx.routers.job_manager.access_policies:WMSAccessPolicy"
 SandboxAccessPolicy = "diracx.routers.job_manager.access_policies:SandboxAccessPolicy"
 
+[project.entry-points."diracx.min_client_version"]
+diracx = "diracx.routers:DIRACX_MIN_CLIENT_VERSION"
 
 [tool.setuptools.packages.find]
 where = ["src"]
--- a/diracx-routers/src/diracx/routers/__init__.py
+++ b/diracx-routers/src/diracx/routers/__init__.py
@@ -12,17 +12,21 @@ import logging
 import os
 from collections.abc import AsyncGenerator
 from functools import partial
+from http import HTTPStatus
+from importlib.metadata import EntryPoint, EntryPoints, entry_points
 from logging import Formatter, StreamHandler
 from typing import Any, Awaitable, Callable, Iterable, Sequence, TypeVar, cast
 
 import dotenv
 from cachetools import TTLCache
-from fastapi import APIRouter, Depends, Request, status
+from fastapi import APIRouter, Depends, FastAPI, HTTPException, Request, status
 from fastapi.dependencies.models import Dependant
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, Response
 from fastapi.routing import APIRoute
+from packaging.version import parse
 from pydantic import TypeAdapter
+from starlette.middleware.base import BaseHTTPMiddleware
 
 # from starlette.types import ASGIApp
 from uvicorn.logging import AccessFormatter, DefaultFormatter
@@ -289,7 +293,6 @@ def create_app_inner(
     origins = [
         "http://localhost:8000",
     ]
-
     app.add_middleware(
         CORSMiddleware,
         allow_origins=origins,
@@ -297,7 +300,7 @@ def create_app_inner(
         allow_methods=["*"],
         allow_headers=["*"],
     )
-
+    app.add_middleware(ClientMinVersionCheckMiddleware)
     configure_logger()
     instrument_otel(app)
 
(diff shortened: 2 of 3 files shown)


natthan-pigoux:
closes #258

natthan-pigoux:
> * Need to check why the `send` or `send_request` client method are not used to send the request.
> * It looks like the request goes through: `client/aio/operations/_operations.py` > `self._client._pipeline.run` when testing with `userinfo`

@chaen find how to add properly the header in the client:
kwargs.setdefault("base_headers", {})[DiracX-Client-Version"] = self.client_version

natthan-pigoux:
Do I need to add more unit tests ?

diracx-routers/pyproject.toml
```diff
@@ -56,6 +56,8 @@ auth = "diracx.routers.auth:router"
 WMSAccessPolicy = "diracx.routers.job_manager.access_policies:WMSAccessPolicy"
 SandboxAccessPolicy = "diracx.routers.job_manager.access_policies:SandboxAccessPolicy"
 
+[project.entry-points."diracx.min_client_version"]
+diracx = "diracx.routers:DIRACX_MIN_CLIENT_VERSION"
```
aldbr, line 60: Any reason not to write directly the version in `pyproject.toml`:

```suggestion
diracx = "0.1.0"
```

It sounds like it would be easier to bump the version from a CI job (e.g. based on the project version) for instance if it is written in `pyproject.toml` rather than in a variable in the code.

aldbr:
> Do I need to add more unit tests ?

- a test when the header is missing? (which might be the case if we interact with the server using the openapi or the web interface)
- a test when the header is present but does not contain any value?
- a test when something other than a version is provided? Making sure the server correctly handles weird strings or numbers and does not crash.

diracx-routers/pyproject.toml
natthan-pigoux: Yes, the reason is that you cannot write string directly as value of an "entry-points". 
From what I understand, if wa want a custom metadata we can use "[tool.diracx.min_client_version]" but in that case we can't retreive the value using "importlib.metadata".

natthan-pigoux:
> > Do I need to add more unit tests ?
> 
> * a test when the header is missing? (which might be the case if we interact with the server using the openapi or the web interface)
> * a test when the header is present but does not contain any value?
> * a test when something other than a version is provided? Making sure the server correctly handles weird strings or numbers and does not crash.

last commits add these tests: https://github.com/DIRACGrid/diracx/pull/267/commits/92fac943408770421973cc0f47e95a54efd9ab72

diracx-routers/src/diracx/routers/__init__.py
```diff
@@ -437,3 +443,60 @@ async def db_transaction(db: T2) -> AsyncGenerator[T2, None]:
...
+            logger.info(iv_exc)
+            return None
+
+
+# I'm not sure if this has to be define here:
+DIRACX_MIN_CLIENT_VERSION = "0.0.1"
+
+
+def get_min_client_version():
+    """Extracting min client version from entry_points and seraching for extension."""
```
aldbr, line 490:
```suggestion
    """Extracting min client version from entry_points and searching for extension."""
```

diracx-routers/src/diracx/routers/__init__.py
```diff
@@ -446,3 +452,50 @@ async def db_transaction(db: T2) -> AsyncGenerator[T2, None]:
...
+
+        response = await call_next(request)
+        return response
+
+    def is_version_too_old(self, client_version: str) -> bool | None:
+        """Verify that client version is ge than min."""
+        try:
+            return parse(client_version) < parse(self.min_client_version)
+        except InvalidVersion as iv_exc:
+            logger.info(iv_exc)
```
aldbr, line 484: This will be printed on server side, the client will not have a meaningful error on its side. Is this expected?
May be you should not catch this error or catch it but raise another `HTTPException` (`Invalid client version`) in this case?
And then you can adjust the test accordingly (make sure it raises an exception instead of capturing the stdout/stderr.

diracx-routers/tests/test_generic.py
```diff
@@ -46,3 +52,19 @@ def test_unavailable_db(monkeypatch, test_client):
...
+    )
+    with pytest.raises(HTTPException) as response:
+        test_client.get("/", headers={"DiracX-Client-Version": lower_version_than_min})
+    assert response.value.status_code == HTTPStatus.UPGRADE_REQUIRED
+    assert "not recent enough" in response.value.detail
+
+
+def test_wrong_client_version(test_client, caplog: pytest.LogCaptureFixture):
+    test_client.get("/", headers={"DiracX-Client-Version": "Unknown"})
+    assert "Invalid version: 'Unknown'" in caplog.text
```
aldbr, line 70: This test should be adjusted based on my previous command I think.

diracx-routers/src/diracx/routers/__init__.py
```diff
@@ -440,3 +446,46 @@ async def db_transaction(db: T2) -> AsyncGenerator[T2]:
...
+                status_code=HTTPStatus.UPGRADE_REQUIRED,
+                detail=f"Client version ({client_version}) not recent enough (>= {self.min_client_version}). Upgrade.",
+            )
+
+        response = await call_next(request)
+        return response
+
+    def is_version_too_old(self, client_version: str) -> bool | None:
+        """Verify that client version is ge than min."""
+        return parse(client_version) < parse(self.min_client_version)
```
aldbr, line 475: Actually if you don't catch the `InvalidVersion` exception, FastAPI returns a `500` error to the user, which is not what you want actually.

What about?

```suggestion
    def is_version_too_old(self, client_version: str) -> bool:
        """Verify that client version is ge than min."""
        try:
            return parse(client_version) < self.parsed_min_client_version)
        except InvalidVersion:
            raise HTTPException(
                status_code=400,
                detail=f"Invalid version string: '{version_str}'"
            )
```

Note1: the return type is `bool` in my example, because the function raises an `HTTPException` if there is an issue with the version.

Note2: `parse(self.min_client_version)` is replaced by `self.parsed_min_client_version` for 2 reasons. (i) even though the `parse` function is cheap, it is executed on every request, so it is just a very minor optimization; (ii) if for some reason there is an issue with `min_client_version` (e.g. typo like `1.2]4`), it would raise a `500` error, which is correct in this case, because it would come from the server configuration, and thus we avoid any ambiguity.

diracx-routers/src/diracx/routers/__init__.py
```diff
@@ -440,3 +446,46 @@ async def db_transaction(db: T2) -> AsyncGenerator[T2]:
...
+
+
+class ClientMinVersionCheckMiddleware(BaseHTTPMiddleware):
+    """Custom FastAPI middleware to verify that
+    the client has the required minimum version.
+    """
+
+    def __init__(self, app: FastAPI):
+        super().__init__(app)
+        self.min_client_version = get_min_client_version()
```
aldbr, line 458: Based on `Note2` in my previous comment:

```suggestion
        self.min_client_version = get_min_client_version()
        self.parsed_min_client_version = parse(self.min_client_version)
```

aldbr: Here I guess you don't need to catch it because it will be displayed on server side for the operators that misconfigured DiracX. What do you think?

natthan-pigoux: Yes, I agree!

natthan-pigoux:
Please tell me when review is ok, so that I can squash the commits.

aldbr:
Alright, you can squash the commits, it looks good :+1:

aldbr (approved):

diracx-routers/src/diracx/routers/version.py
```diff
@@ -0,0 +1 @@
+DIRACX_MIN_CLIENT_VERSION = "0.0.1"
```
chaen, line 1: I'd prefer to have that constant in `routers/__init__.py`.  `routers.version` is too misleading as a name I believe

diracx-routers/src/diracx/routers/__init__.py
```diff
@@ -442,3 +448,53 @@ async def db_transaction(db: T2) -> AsyncGenerator[T2]:
...
+        except InvalidVersion as iv_exc:
+            raise HTTPException(
+                status_code=HTTPStatus.BAD_REQUEST,
+                detail=f"Invalid version string: '{client_version}'",
+            ) from iv_exc
+
+
+def get_min_client_version():
+    """Extracting min client version from entry_points and searching for extension."""
+    matched_entry_points: EntryPoints = entry_points(group="diracx.min_client_version")
```
chaen, line 489: This just made me realize that it is only compatible with versioning the vanilla diracx client, and not the extension.  You already spent a substantial amount of time on this PR, so I will not ask you to make it extension-compatible :laughing:  I'll just open an issue with it

chaen (approved):

PR #269, opened 2024-07-01
feat: enable remote pilot logging system [MISSING AUTH]

martynia:
Enable remote logging system with open search. This PR only shows a router in its preliminary form.
The router defines following operations:

1. `@router.post("/")`  - bulk insert log record sent by a pilot. The method returns a` pilotID`, so it can use it next time to insert messaged w/o consulting `PilotAgentsDB` to get PilotID from `PilotStamp`. 
2. `@router.get("/logs")` - to retrieve logs for a given pilot_id

Policies allow pilots to store VO-aware logs. Normal users can only retrieve log for their own VOs. diracAdmin can retrieve logs for any VO.

### DIFF
--- a/diracx-db/src/diracx/db/os/utils.py
+++ b/diracx-db/src/diracx/db/os/utils.py
@@ -13,6 +13,7 @@ from datetime import datetime
 from typing import Any, Self
 
 from opensearchpy import AsyncOpenSearch
+from opensearchpy.helpers import async_bulk
 
 from diracx.core.exceptions import InvalidQueryError
 from diracx.core.extensions import select_from_extension
@@ -192,6 +193,13 @@ class BaseOSDB(metaclass=ABCMeta):
         )
         print(f"{response=}")
 
+    async def bulk_insert(self, index_name: str, docs: list[dict[str, Any]]) -> None:
+        """Bulk inserting to database."""
+        n_inserted = await async_bulk(
+            self.client, actions=[doc | {"_index": index_name} for doc in docs]
+        )
+        logger.info("Inserted %d documents to %r", n_inserted, index_name)
+
     async def search(
         self, parameters, search, sorts, *, per_page: int = 100, page: int | None = None
     ) -> list[dict[str, Any]]:
@@ -232,6 +240,17 @@ class BaseOSDB(metaclass=ABCMeta):
                     )
 
         return hits
+
+    async def delete(self, query: list[dict[str, Any]]) -> dict:
+        """Delete multiple documents by query."""
+        body = {}
+        res = {}
+        if query:
+            body["query"] = apply_search_filters(self.fields, query)
+            res = await self.client.delete_by_query(
+                body=body, index=f"{self.index_prefix}*"
+            )
+        return res
 
 
 def require_type(operator, field_name, field_type, allowed_types):
--- a/diracx-routers/src/diracx/routers/pilots/logging.py
+++ b/diracx-routers/src/diracx/routers/pilots/logging.py
@@ -3,9 +3,10 @@ from __future__ import annotations
 import logging
 
 from fastapi import HTTPException, status
-from pydantic import BaseModel
 
-from diracx.core.models import ScalarSearchOperator, ScalarSearchSpec
+from diracx.core.models import LogMessage
+from diracx.logic.pilots.logging import get_logs as get_logs_bl
+from diracx.logic.pilots.logging import send_message as send_message_bl
 
 from ..access_policies import open_access
 from ..dependencies import PilotAgentsDB, PilotLogsDB
@@ -14,22 +15,6 @@ from .access_policies import ActionType, CheckPilotLogsPolicyCallable
 
 logger = logging.getLogger(__name__)
 router = DiracxRouter()
-
-
-class LogLine(BaseModel):
-    line_no: int
-    line: str
-
-
-class LogMessage(BaseModel):
-    pilot_stamp: str
-    lines: list[LogLine]
-    vo: str
-
-
-class DateRange(BaseModel):
-    min: str | None = None  # expects a string in ISO 8601 ("%Y-%m-%dT%H:%M:%S.%f%z")
-    max: str | None = None  # expects a string in ISO 8601 ("%Y-%m-%dT%H:%M:%S.%f%z")
 
 
 @open_access
@@ -41,55 +26,11 @@ async def send_message(
     # user_info: Annotated[AuthorizedUserInfo, Depends(verify_dirac_access_token)],
 ) -> int:
 
-    # expecting exactly one row:
-    search_params = ScalarSearchSpec(
-        parameter="PilotStamp",
-        operator=ScalarSearchOperator.EQUAL,
-        value=data.pilot_stamp,
-    )
-
-    total, result = await pilot_agents_db.search(
-        ["PilotID", "VO", "SubmissionTime"], [search_params], []
-    )
-    if total != 1:
-        logger.error(
-            "Cannot determine PilotID for requested PilotStamp: %r, (%d candidates)",
-            data.pilot_stamp,
-            total,
-        )
-        raise HTTPException(
-            status.HTTP_400_BAD_REQUEST, detail=f"Number of rows !=1: {total}"
-        )
-    pilot_id, vo, submission_time = (
-        result[0]["PilotID"],
-        result[0]["VO"],
-        result[0]["SubmissionTime"],
-    )
-
-    # await check_permissions(action=ActionType.CREATE, pilot_agent_db, pilot_id),
-
-    docs = []
-    for line in data.lines:
-        docs.append(
-            {
-                "PilotStamp": data.pilot_stamp,
-                "PilotID": pilot_id,
-                "SubmissionTime": submission_time,
-                "VO": vo,
-                "LineNumber": line.line_no,
-                "Message": line.line,
-            }
-        )
-    await pilot_logs_db.bulk_insert(pilot_logs_db.index_name(pilot_id), docs)
-    """
-    search_params = [{"parameter": "PilotID", "operator": "eq", "value": pilot_id}]
-
-    result = await pilot_logs_db.search(
-        ["Message"],
-        search_params,
-        [{"parameter": "LineNumber", "direction": "asc"}],
-    )
-    """
+    # await check_permissions(action=ActionType.CREATE, pilot_agent_db, pilot_id)
+    try:
+        pilot_id = await send_message_bl(data, pilot_logs_db, pilot_agents_db)
+    except Exception as exc:
+        raise HTTPException(status.HTTP_400_BAD_REQUEST, detail=str(exc)) from exc
     return pilot_id
 
 
@@ -107,13 +48,4 @@ async def get_logs(
         action=ActionType.QUERY, pilot_agents_db=pilot_agents_db, pilot_id=pilot_id
     )
 
-    search_params = [{"parameter": "PilotID", "operator": "eq", "value": pilot_id}]
-
-    result = await db.search(
-        ["Message"],
-        search_params,
-        [{"parameter": "LineNumber", "direction": "asc"}],
-    )
-    if not result:
-        return [{"Message": f"No logs for pilot ID = {pilot_id}"}]
-    return result
+    return await get_logs_bl(pilot_id, db)
--- a/diracx-routers/src/diracx/routers/pilots/logging.py
+++ b/diracx-routers/src/diracx/routers/pilots/logging.py
@@ -1,0 +1,119 @@
+from __future__ import annotations
+
+import logging
+
+from fastapi import HTTPException, status
+from pydantic import BaseModel
+
+from diracx.core.models import ScalarSearchOperator, ScalarSearchSpec
+
+from ..access_policies import open_access
+from ..dependencies import PilotAgentsDB, PilotLogsDB
+from ..fastapi_classes import DiracxRouter
+from .access_policies import ActionType, CheckPilotLogsPolicyCallable
+
+logger = logging.getLogger(__name__)
+router = DiracxRouter()
+
+
+class LogLine(BaseModel):
+    line_no: int
+    line: str
+
+
+class LogMessage(BaseModel):
+    pilot_stamp: str
+    lines: list[LogLine]
+    vo: str
+
+
+class DateRange(BaseModel):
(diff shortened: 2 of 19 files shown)


martynia:
@chrisburr I think we need a pilot login router, even in its basic form. One of the tests fails because it is missing.

martynia:
@chrisburr What is a reason of test failures here ? The demo runs fine locally, can be contacted with a client etc. The test fails with `dirac login`, waiting for `diracX` to be installed. Could it be that I missed a setting somewhere ? The failure is persistent.

martynia:
@chrisburr @chaen 
I managed to reproduce the problem locally. There is the same failure on GitHub and locally. Since the demo runs fine (the DB is there), is it possible that integration tests fail `(NotImplementedError: Cannot enable system_name='pilotlogs' as it requires missing_os_dbs={<class 'diracx.db.os.pilot_logs.PilotLogsDB'>}` because of charts being no discoverable by the CI workflow ?

This type of error occurs (in the demo) where charts don't list a relevant DB

Both local and CI tests reach the same point - server installation failure.

chrisburr:
https://github.com/DIRACGrid/DIRAC/blob/7fe654eef5f4b0c23a7047e6e046580b1e354e97/tests/CI/docker-compose.yml#L174-L209

martynia:
@chaen I fixed the database problem, now integration tests are passing, however I'm getting gubbins errors in unit tests. Is it an internal gubbins problem ?

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
```diff
@@ -0,0 +1,156 @@
...
+        search_params.append(
+            {"parameter": "VO", "operator": "eq", "value": user_info.vo}
+        )
+    result = await db.search(
+        ["Message"],
+        search_params,
+        [{"parameter": "LineNumber", "direction": "asc"}],
+    )
+    if not result:
+        return [{"Message": f"No logs for pilot ID = {pilot_id}"}]
```
aldbr, line 108: Similarly, this bloc of code should be a method of `diracx-dbs/src/diracx/db/sql/pilots/pilot_agents.py PilotAgentsDB`

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
```diff
@@ -0,0 +1,156 @@
...
+async def delete(
+    pilot_id: int,
+    data: DateRange,
+    db: PilotLogsDB,
+    check_permissions: CheckPilotLogsPolicyCallable,
+) -> str:
+    """Delete either logs for a specific PilotID or a creation date range.
+    Non-privileged users can only delete log files within their own VO.
+    """
+    message = "no-op"
```
aldbr, line 122: Why do you initialize `message` as `"no-op"`?

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
```diff
@@ -0,0 +1,156 @@
...
+        message = f"Logs for for pilots with submission data >='{data.min}' successfully deleted"
+
+    return message
+
+
+def _non_privileged(user_info: AuthorizedUserInfo):
+    return (
+        SERVICE_ADMINISTRATOR not in user_info.properties
+        and OPERATOR not in user_info.properties
+    )
```
aldbr, line 156: I feel like this should be moved in the `access_policies.py` function? May be by adding an action type? Or by adding some logic like in:
 https://github.com/DIRACGrid/diracx/blob/320c554934eb8d81828c94f89608c5347f7edd0a/diracx-routers/src/diracx/routers/jobs/access_policies.py#L79-L82

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
```diff
@@ -0,0 +1,156 @@
...
+        [{"parameter": "LineNumber", "direction": "asc"}],
+    )
+    if not result:
+        return [{"Message": f"No logs for pilot ID = {pilot_id}"}]
+    return result
+
+
+@router.delete("/logs")
+async def delete(
+    pilot_id: int,
```
aldbr, line 114: We also tend to make bulk requests to the db to improve the performances of the system. I would take a list of `pilot_ids` as inputs instead of a single `pilot_id`.

Then you can return `sucess` and `failures` as it is done for the jobs for instance:

https://github.com/DIRACGrid/diracx/blob/320c554934eb8d81828c94f89608c5347f7edd0a/diracx-routers/src/diracx/routers/jobs/status.py#L35-L62

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
```diff
@@ -0,0 +1,156 @@
...
+                "Message": line.line,
+            }
+        )
+    await pilot_logs_db.bulk_insert(pilot_logs_db.index_name(pilot_id), docs)
+    return pilot_id
+
+
+@router.get("/logs")
+async def get_logs(
+    pilot_id: int,
```
aldbr, line 88: Similarly, 
```suggestion
    pilot_ids: List[int],
```

aldbr (changes requested):

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
martynia: `sql/` ? 

This is an `os` search.

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
martynia: Well, this is a return message to the client. As far as I can see, the delete by query operation is silent. It does not complain when you delete something which doesn't not exist etc. Or I might have missed something in the docs.

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
martynia: I had a logical problem what to do here. I would like to search only once (and delete by query for a delete operation w/o prior searching). This is a reason why I removed the DB argument from a policy call. The problem  is more visible in the delete case:

- I could search in the policy for a given client VO and if pilot IDs are not from the VO throw an exception (not sure what to do if some IDs are and some are not from the VO the client has - for a bulk PilotID search)
- then delete by query (which searches again)

Or could I use the AgentsDB to match the VO ? Still an extra DB query, this time and SQL one.
I opted for a restricted silent delete and and a restricted search by adding the VO, but outside the policy.

diracx-routers/src/diracx/routers/pilot_logging/__init__.py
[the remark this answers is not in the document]
martynia: OK, could do this.

diracx-db/src/diracx/db/os/utils.py
```diff
@@ -231,6 +239,15 @@ async def search(
 
         return hits
 
+    async def delete(self, query: list[dict[str, Any]]) -> None:
+
+        # Delete multiple documents by query.
+
```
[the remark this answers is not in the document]
martynia: Done

diracx-routers/src/diracx/routers/pilot_logging/__init__.py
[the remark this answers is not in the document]
martynia: Done

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
```diff
@@ -0,0 +1,156 @@
...
+
+class LogMessage(BaseModel):
+    pilot_stamp: str
+    lines: list[LogLine]
+    vo: str
+
+
+class DateRange(BaseModel):
+    min: str | None = None  # expects a string in ISO 8601 ("%Y-%m-%dT%H:%M:%S.%f%z")
+    max: str | None = None  # expects a string in ISO 8601 ("%Y-%m-%dT%H:%M:%S.%f%z")
```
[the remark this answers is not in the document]
aldbr: Also you can probably check that `min` < `max` using pydantic

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
```diff
@@ -0,0 +1,156 @@
...
+            async with db:
+                stmt = select(PilotAgents.PilotID, PilotAgents.SubmissionTime).where(
+                    PilotAgents.PilotStamp == data.pilot_stamp
+                )
+                pilot_id, submission_time = (await db.conn.execute(stmt)).one()
+    except NoResultFound as exc:
+        logger.error(
+            f"Cannot determine PilotID for requested PilotStamp: {data.pilot_stamp}, Error: {exc}."
+        )
+        raise HTTPException(status.HTTP_400_BAD_REQUEST, detail=str(exc)) from exc
```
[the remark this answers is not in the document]
aldbr: FYI, I am removing `sqlalchemy` dependency from the `diracx-routers` to prevent writing data access layer code in the routers: https://github.com/DIRACGrid/diracx/issues/358

If my PR is merged and you don't move that bloc of code to `diracx-db`, tests will likely fail.

diracx-routers/src/diracx/routers/pilots/logging.py
```diff
@@ -0,0 +1,159 @@
...
+
+    elif data.min:
+        logger.warning(f"Deleting logs for pilots with submission data >='{data.min}'")
+        search_params = [
+            {"parameter": "SubmissionTime", "operator": "gt", "value": data.min}
+        ]
+        if _non_privileged(user_info):
+            search_params.append(non_privil_params)
+        await db.delete(search_params)
+        message = f"Logs for for pilots with submission data >='{data.min}' successfully deleted"
```
aldbr, line 150:
```suggestion
        message = f"Logs for pilots with submission data >='{data.min}' successfully deleted"
```

diracx-routers/src/diracx/routers/pilots/logging.py
```diff
@@ -0,0 +1,159 @@
...
+) -> str:
+    """Delete either logs for a specific PilotID or a creation date range.
+    Non-privileged users can only delete log files within their own VO.
+    """
+    message = "no-op"
+    user_info = await check_permissions(action=ActionType.DELETE)
+    non_privil_params = {"parameter": "VO", "operator": "eq", "value": user_info.vo}
+
+    # id pilot_id is provided we ignore data.min and data.max
+    if not pilot_ids and data.min and data.max:
```
aldbr, line 128: May be I didn't read carefully, but don't you mean:
```suggestion
    if not pilot_ids and not (data.min or data.max):
```

diracx-routers/src/diracx/routers/pilots/logging.py
```diff
@@ -0,0 +1,159 @@
...
+        search_params,
+        [{"parameter": "LineNumber", "direction": "asc"}],
+    )
+    if not result:
+        return [{"Message": f"No logs for pilot ID = {pilot_id}"}]
+    return result
+
+
+@router.delete("/logs")
+async def delete(
```
aldbr, line 114: By the way, I am just curious, what is the use case of `DELETE /pilots/logs`?

diracx-routers/src/diracx/routers/pilot_logging/remote_logger.py
aldbr: > Well, this is a return message to the client.

But IIUC, this `"no-op"` can't be returned to the client, right?
Then may be you can just initialize it to `None` instead of `"no-op"`?

> As far as I can see, the delete by query operation is silent. It does not complain when you delete something which doesn't not exist etc. Or I might have missed something in the docs.

You mean that if you pass a wrong pilot id or a wrong date to OpenSearch, it does not inform you whether pilot ids and/or dates are all valid?

I am not an expert of opensearch, I don't have much knowledge about it, but from what I can read,  https://opensearch.org/docs/latest/api-reference/document-apis/delete-by-query/ seems to provide the number of deleted items, but not the details (which ones).

One solution to identify them could consist in searching for the pilot ids first,  and then return the states of each one after the delete query. 

Something like (this is pseudo-code):

```python
pilots_ids: [1,2,3,4,5]

result = search(pilots_ids)
# result = [1,2,3]

unknown_pilot_ids = set(pilot_ids) - set(result)
# unkown_pilots_ids = [4,5]

delete_by_query(result)

return {"success": result ,"failed": unknown_pilot_ids}
```

diracx-routers/src/diracx/routers/pilots/logging.py
martynia: to delete old logs ?

diracx-routers/src/diracx/routers/pilots/logging.py
martynia: The logic is correct. `pilot_id`has priority, if it is not given then we have to cases:

1. a range is given both `min` and `max` (this requires a (simple) change in DiracX query language - open search support range out of the box; also open range.)
2. only `max` is given, then delete everything olde. that `max.`

diracx-routers/src/diracx/routers/pilots/logging.py
martynia: Sure.

diracx-routers/src/diracx/routers/pilots/logging.py
```diff
@@ -0,0 +1,162 @@
...
+async def get_logs(
+    pilot_id: int,
+    db: PilotLogsDB,
+    check_permissions: CheckPilotLogsPolicyCallable,
+) -> list[dict]:
+
+    logger.warning(f"Retrieving logs for pilot ID '{pilot_id}'")
+    user_info = await check_permissions(action=ActionType.QUERY)
+
+    # here, users with privileged properties will see logs from all VOs. Is it what we want ?
```
fstagni, line 97: These are hardly secrets

diracx-routers/src/diracx/routers/pilots/logging.py
martynia: Removed.

diracx-routers/src/diracx/routers/pilots/logging.py
martynia: Irrelevant now, since the DELETE operation has been deleted.

chaen:
Waiting for the pilot authentication to be ready

martynia:
move to Pilot logging PR

Issue #89, opened 2023-09-15
Test SQL transaction model


chrisburr:
@chaen will discuss with @aldbr 

Should handle:

* The normal "transparent" transaction with automatic commit/rollback in a request
* Special cases like the refresh token revocation where we want to commit while also returning an error
* Document how to handle the DB transactions

PR #271, opened 2024-07-02
feat: add SQL transaction model tests

aldbr:
Closes https://github.com/DIRACGrid/diracx/issues/89

There are a few details about the transaction model here: https://github.com/DIRACGrid/diracx/blob/ea076d7e6434c80ccb1c798c465c81f2ecf4141a/docs/SERVICES.md?plain=1#L96

Would you like more details?


chaen:
If possible could you please add an "advanced guide" for when you want to commit within the DB anyway (like what you did in fact)

chrisburr (approved):

PR #272, opened 2024-07-05
Align python versions, add pilot tests to legacy tests

fstagni:


fstagni:
This is ending up in a mess. I will separate the various commits in at least 2 PRs.

Issue #264, opened 2024-06-24
Try to setup release-please (or alternatives ... semantic-releases?)


aldbr:
Based on the work done in https://github.com/DIRACGrid/diracx-web/pull/134

fstagni:
Requirements:
- should work for diracx, web, charts
- should work on github and gitlab

chrisburr:
Probably want to use: https://github.com/semantic-release/semantic-release

PR #281, opened 2024-08-21
Fixes looking around

fstagni:


diracx-routers/tests/test_job_manager.py
```diff
@@ -569,7 +569,7 @@ def test_set_job_status_offset_naive_datetime_return_bad_request(
     valid_job_id: int,
 ):
     # Act
-    date = datetime.utcnow().isoformat(sep=" ")  # noqa: DTZ003
```
chaen, line 572: I think this was on purpose, given that there was the proper `noqa` https://docs.astral.sh/ruff/rules/call-datetime-utcnow/
I would guess it has to do with DIRAC compatibility but I don't remember

fstagni: `utcnow` will be deprecated. What I put is of course equivalent.

.github/workflows/main.yml
```diff
@@ -14,6 +14,34 @@ defaults:
     shell: bash -el {0}
 
 jobs:
+
+  pre-commit:
```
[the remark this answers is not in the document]
fstagni: That's true. I can remove indeed this check.

diracx-routers/tests/test_job_manager.py
chaen: Even though it is subtle, it is not

```python
In [19]: tz_now = datetime.now(tz=timezone.utc); now = datetime.utcnow()

In [20]: tz_now
Out[20]: datetime.datetime(2024, 8, 22, 21, 13, 40, 232387, tzinfo=datetime.timezone.utc)

In [21]: now
Out[21]: datetime.datetime(2024, 8, 22, 21, 13, 40, 232433)

In [22]: tz_now - now
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[22], line 1
----> 1 tz_now - now

TypeError: can't subtract offset-naive and offset-aware datetimes
```

I was tricked by that in DIRAC in the past:

https://github.com/DIRACGrid/DIRAC/blob/f18be6f9d27f5055512f2c372718734270a060ae/src/DIRAC/Core/Security/m2crypto/asn1_utils.py#L196-L201

But indeed here we should check why was this done and maybe your fix is good.

chaen:
Why removing the `__future__` import ?

diracx-routers/tests/test_job_manager.py
fstagni: Further converting it is OK:

```python
In [9]: formatted__tz_now = tz_now.isoformat(sep=' ').split("+")[0]

In [10]: formatted__now = now.isoformat(sep=' ').split("+")[0]

In [14]: formatted__tz_now
Out[14]: '2024-08-23 08:56:13.234331'

In [15]: formatted__now
Out[15]: '2024-08-23 08:56:13.234372'

In [16]: date_obj__tz_now = datetime.strptime(formatted__tz_now, '%Y-%m-%d %H:%M:%S.%f')

In [17]: date_obj__now = datetime.strptime(formatted__now, '%Y-%m-%d %H:%M:%S.%f')

In [18]: date_obj__tz_now -  date_obj__now
Out[18]: datetime.timedelta(days=-1, seconds=86399, microseconds=999959)

```

But:

```
In [20]: date_obj__tz_now__no_split = datetime.strptime(formatted__tz_now__no_split, '%Y-%m-%d %H:%M:%S.%f')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[20], line 1
----> 1 date_obj__tz_now__no_split = datetime.strptime(formatted__tz_now__no_split, '%Y-%m-%d %H:%M:%S.%f')

File ~/miniforge3/envs/dirac-development/lib/python3.11/_strptime.py:567, in _strptime_datetime(cls, data_string, format)
    564 def _strptime_datetime(cls, data_string, format="%a %b %d %H:%M:%S %Y"):
    565     """Return a class cls instance based on the input string and the
    566     format string."""
--> 567     tt, fraction, gmtoff_fraction = _strptime(data_string, format)
    568     tzname, gmtoff = tt[-2:]
    569     args = tt[:6] + (fraction,)

File ~/miniforge3/envs/dirac-development/lib/python3.11/_strptime.py:352, in _strptime(data_string, format)
    349     raise ValueError("time data %r does not match format %r" %
    350                      (data_string, format))
    351 if len(data_string) != found.end():
--> 352     raise ValueError("unconverted data remains: %s" %
    353                       data_string[found.end():])
    355 iso_year = year = None
    356 month = day = 1

ValueError: unconverted data remains: +00:00

In [21]: formatted__tz_now__no_split
Out[21]: '2024-08-23 08:56:13.234331+00:00'

In [22]: date_obj__tz_now__no_split = datetime.strptime(formatted__tz_now__no_split, '%Y-%m-%d %H:%M:%S.%f+00:00')

In [23]:  date_obj__tz_now__no_split
Out[23]: datetime.datetime(2024, 8, 23, 8, 56, 13, 234331)

In [24]: date_obj__tz_now__no_split -  date_obj__now
Out[24]: datetime.timedelta(days=-1, seconds=86399, microseconds=999959)
```

...I guess it is a feature, but to me it even feels like a python bug! Bof!

fstagni:
> Why removing the `__future__` import ?

Because since python 3.7 is only necessary for specific cases.

chrisburr:
> Because since python 3.7 is only necessary for specific cases.

It was added in 3.7 so that can't be the case :wink:

The intention is to make the `from __future__ import annotations` (see [PEP-563](https://peps.python.org/pep-0563/) but it keeps getting deffered.

chaen:
It actually should be everywhere except specific places like https://github.com/DIRACGrid/diracx/blob/b86ed1180bd05011d5aef8f8327d8c3d2f0a7c34/diracx-routers/src/diracx/routers/access_policies.py#L31-L35

fstagni:
Sorry, I meant 3.10, but indeed I realise now that the decision to make it mandatory was postponed and it is not even in 3.12. So, I revert this change.

.github/workflows/semantic.yml
```diff
@@ -0,0 +1,33 @@
...
+# and that they follow the rule
+# <type> (<scope>): <subject>
+# type: docs, feat, fix, refactor, style or test
+# scope (optional): any extra info, (like DMS or whatever)
+
+name: 'Commit Message Check'
+on: pull_request
+
+jobs:
+  check-commit-message:
```
chrisburr, line 11: The reason we didn't have this was that it can probably be done with pre-commit.

Also, this PR feels like many merged into one.

.github/workflows/main.yml
```diff
@@ -14,6 +14,21 @@ defaults:
     shell: bash -el {0}
 
 jobs:
+
+  shellcheck:
```
chrisburr, line 18: Do we have any shell scripts to check?

diracx-core/tests/test_secrets.py
```diff
@@ -15,8 +15,7 @@ def compare_keys(key1, key2):
 def test_token_signing_key(tmp_path):
     private_key = rsa.generate_private_key(
         public_exponent=65537,
-        # DANGER: 512-bits is a bad idea for prod but makes the test notably faster!
-        key_size=512,
+        key_size=1024,
```
chrisburr, line 18: Can we avoid making the tests slower by using elliptic curves?

diracx-routers/tests/jobs/test_sandboxes.py
```diff
@@ -34,7 +34,7 @@ def normal_user_client(client_factory):
 def test_upload_then_download(
     normal_user_client: TestClient, test_auth_settings: AuthSettings
 ):
-    data = secrets.token_bytes(512)
+    data = secrets.token_bytes(1024)
```
chrisburr, line 37: Why?

.github/workflows/main.yml
fstagni: Only 1 for the moment (with some fixes from this same check).

.github/workflows/semantic.yml
chaen: it's actually going to be driven by https://github.com/DIRACGrid/diracx/issues/264

.github/workflows/main.yml
```diff
@@ -39,13 +54,7 @@ jobs:
...
         with:
           # TODO: Use a conda environment file used for the diracx/base container image
-          environment-name: test-env
-          create-args: >-
-            python=3.11
-            m2crypto
-            python-gfal2
-            mypy
-            pip
+          environment-file: environment.yml
```
chaen, line 57: it's not this `environment.yml` but the one from `container-images` that should be used

.github/workflows/semantic.yml
fstagni: Correct, removed.

diracx-core/tests/test_secrets.py
fstagni: In this PR? Can we leave for another one?

.github/workflows/main.yml
fstagni: This `environment.yml` defines the environment where all of us are working locally, right? So, it's the natural environment where all unit tests run. 

In `container-images` there are many `environment.yml` files. What is here is, permit me that, the superset of all of them.

chaen: absolutely, but this is what we want to try to avoid. We want to have the minimal set of dependencies installed to keep control of what we *actually* need, and this is why we did not blindly put the `environment.yml` before

fstagni: So, you want to repeat, for ONLY the unit tests run by the CI, the ~same environment (because it is not going to be very different) that we all have in our local machines, and that we all use for running the unit tests locally? Meaning that I can run the unit tests locally, they all pass, then create the PR only to discover that there they are failing. If there's a version to restrict (like for the one for typer), that, like it or not, we have to do it from time to time, then we would need to apply that fix in even one more place.

There is already a [noted duplication between setup.cfg and environment.yml](https://github.com/DIRACGrid/diracx/blob/main/docs/VERSIONING.md#dependencies) and I do not think we should have even more.

The environment where the unit tests run is already NOT what it is run in production. Which is fair and fine. But that is why the integrations tests are there.

chrisburr: We expect the CI to run in an environment which is different to out local machines as we intetionally only install the bare minimum in the CI. We also let it use the newest possible version to try and catch issues before they affect the harder/slower to debug tests.

Maybe we want to drop the environment.yaml file entirely but that's a separate discussion to start and maybe we want to use [something else to harmonise this](https://pixi.sh/latest/features/multi_environment/) but we've not got around to trying it.

.pre-commit-config.yaml
```diff
@@ -36,3 +36,10 @@ repos:
           - types-aiobotocore[essential]
           - boto3-stubs[essential]
         exclude: ^(diracx-client/src/diracx/client/|diracx-[a-z]+/tests/|diracx-testing/|build)
+
+  - repo: https://github.com/asottile/pyupgrade
```
chrisburr, line 40: We have discussed this before...

chrisburr: https://github.com/DIRACGrid/diracx/pull/61#discussion_r1311558741

fstagni:
Superseded by #286

PR #285, opened 2024-08-28
Fix nested access policy and violent crash

### LINKED ISSUE #275: AccessPolicy checks hide useful errors
If you are calling a method with wrong arguments, you are supposed to see an error 402 with what is wrong.
Instead, you get nothing because the server crashes with the infamous `"THIS SHOULD NOT HAPPEN, ALWAYS VERIFY PERMISSION",`

### LINKED ISSUE #283: Nested access policies is broken: cannot inject the inner access policy (`SandboxAccessPolicy` with `WMSAccessPolicy`)
**How to reproduce**

- run the demo and log in
- submit a job
- try to get the sandbox (`GET /api/jobs/sandbox`)

FastAPI returns `Error 500`. Logs are:

```bash
  File "/diracx_source/diracx/diracx-routers/src/diracx/routers/job_manager/sandboxes.py", line 228, in get_job_sandboxes
    await check_permissions(action=ActionType.READ, job_db=job_db, job_ids=[job_id])
  File "/diracx_source/diracx/diracx-routers/src/diracx/routers/access_policies.py", line 118, in wrapped_policy
    return await policy(policy_name, user_info, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/diracx_source/diracx/diracx-routers/src/diracx/routers/job_manager/access_policies.py", line 127, in policy
    assert check_wms_permissions
           ^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

**Problem**

Not entirely sure as the access policy mechanism is not easy to understand, but IIUC, `CheckWMSPolicyCallable` is neither a direct dependency nor a subdependency (nested dependency) of the sandbox routes, and there is no way of defining dependencies (`Depends`) outside of `fastAPI` context. 

**Possible solutions**

- 1: we don't make nested access policies(e.g. we use the `WMSAccessPolicy` instead of the `SandboxAccessPolicy` when required: when `job_ids` or `job_db` are part of the arguments of the route).
- 2: we modify the access policy mechanism, but so far I haven't found any simple fix.

Any thought?

chaen:
Closes #275 
Closes #283

### DIFF
--- a/diracx-routers/src/diracx/routers/__init__.py
+++ b/diracx-routers/src/diracx/routers/__init__.py
@@ -13,7 +13,15 @@ import os
 from collections.abc import AsyncGenerator
 from functools import partial
 from logging import Formatter, StreamHandler
-from typing import Any, Awaitable, Callable, Iterable, Sequence, TypeVar, cast
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Iterable,
+    Sequence,
+    TypeVar,
+    cast,
+)
 
 import dotenv
 from cachetools import TTLCache
@@ -139,6 +147,7 @@ def create_app_inner(
     # Please see ServiceSettingsBase for more details
 
     available_settings_classes: set[type[ServiceSettingsBase]] = set()
+
     for service_settings in all_service_settings:
         cls = type(service_settings)
         assert cls not in available_settings_classes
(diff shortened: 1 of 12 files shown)


diracx-routers/src/diracx/routers/__init__.py
```diff
@@ -139,6 +147,7 @@ def create_app_inner(
     # Please see ServiceSettingsBase for more details
 
     available_settings_classes: set[type[ServiceSettingsBase]] = set()
+    print(f"CHRIS {all_service_settings=}")
```
fstagni, line 150: ...I guess this can be removed.

aldbr (approved):

PR #286, opened 2024-08-28
Fixes looking around

fstagni:


fstagni:
Alright?

.github/workflows/main.yml
```diff
@@ -14,6 +14,21 @@ defaults:
     shell: bash -el {0}
 
 jobs:
+
+  shellcheck:
```
chrisburr, line 18: Is this worth having? I think we only have the run-local.sh file

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 default_language_version:
-  python: python3
+  python: python3.11
```
chrisburr, line 4: I think we should avoid specifying the version here so it's one less place to keep track of

```suggestion
  python: python3
```

.github/workflows/main.yml
fstagni: It does not hurt

fstagni:
OK now?

chrisburr (approved):

PR #292, opened 2024-09-11
feat: Implement PilotAgents schema

martynia:
Add PilotAgents related tables. Add PilotAgentsDB to `pyproject.toml`.


diracx-db/src/diracx/db/sql/pilot_agents/schema.py
```diff
@@ -0,0 +1,65 @@
...
+    GridType = Column("GridType", String(32), default="LCG")
+    GridRequirements = Column("GridRequirements", Text, default="")
+    BenchMark = Column("BenchMark", Double, default=0.0)
+    SubmissionTime = NullColumn("SubmissionTime", DateTime)
+    LastUpdateTime = NullColumn("LastUpdateTime", DateTime)
+    Status = Column("Status", String(32), default="Unknown")
+    StatusReason = Column("StatusReason", String(255), default="Unknown")
+    ParentID = Column("ParentID", Integer, default=0)
+    OutputReady = Column("OutputReady", EnumBackedBool(), default=False)
+    AccountingSent = Column("AccountingSent", EnumBackedBool(), default=False)
```
aldbr, line 41: Hi @martynia, I think you copied the DB schema from Dirac `v8.0` here, right?
There have been some simplifications from `v9.0` as you can see here: 

https://github.com/DIRACGrid/DIRAC/blob/826762853ecdd2011bfc9dffa6fef2eeb9842865/src/DIRAC/WorkloadManagementSystem/DB/PilotAgentsDB.sql#L28-L49

martynia: Thanks for the comment. Fixed.

diracx-db/src/diracx/db/sql/pilot_agents/db.py
```diff
@@ -0,0 +1,42 @@
...
+            stmt = insert(PilotAgents).values(
+                PilotJobReference=ref,
+                VO=VO,
+                GridType=gridType,
+                SubmissionTime=now,
+                LastUpdateTime=now,
+                Status="Submitted",
+                PilotStamp=stamp,
+            )
+            result = await self.conn.execute(stmt)
```
aldbr, line 39: I am not an expert with `sqlalchemy`, but I wonder whether this operation should be done for every values to insert.
I think there is a way to insert the all the values in a single db call: https://docs.sqlalchemy.org/en/20/glossary.html#term-executemany

IIUC, here is how you could use it:

```python
        now = datetime.now(tz=timezone.utc)

        # Prepare the list of dictionaries for bulk insertion
        values = [
            {
                "PilotJobReference": ref,
                "VO": vo,
                "GridType": gridType,
                "SubmissionTime": now,
                "LastUpdateTime": now,
                "Status": "Submitted",
                "PilotStamp": pilotStampDict.get(ref, "")
            }
            for ref in pilotRef
        ]

        # Insert multiple rows in a single execute call
        stmt = insert(PilotAgents)
        await self.conn.execute(stmt, values)
```

However, there is one limitation with `executemany` and `mySQL`: you cannot get the rows in return (https://docs.sqlalchemy.org/en/20/core/connections.html#insert-many-values-behavior-for-insert-statements).
I see that you are returning the row ids, but they were not returned in the original method, right? Is there any reason for having them here?

If we don't specifically need to return the ids, it might be worth comparing both methods to see whether there is a huge difference in terms of performances.

diracx-db/src/diracx/db/sql/pilot_agents/db.py
```diff
@@ -0,0 +1,42 @@
...
+from .schema import PilotAgents, PilotAgentsDBBase
+
+
+class PilotAgentsDB(BaseSQLDB):
+    metadata = PilotAgentsDBBase.metadata
+
+    async def addPilotReferences(
+        self,
+        pilotRef: list[str],
+        VO: str,
```
aldbr, line 17:
```suggestion
        vo: str,
```

diracx-db/src/diracx/db/sql/pilot_agents/db.py
```diff
@@ -0,0 +1,42 @@
...
+            pilotStampDict = {}
+        row_ids = []
+        for ref in pilotRef:
+            stamp = ""
+            if ref in pilotStampDict:
+                stamp = pilotStampDict[ref]
+            now = datetime.now(tz=timezone.utc)
+            stmt = insert(PilotAgents).values(
+                PilotJobReference=ref,
+                VO=VO,
```
aldbr, line 32:
```suggestion
                VO=vo,
```

diracx-db/src/diracx/db/sql/pilot_agents/db.py
```diff
@@ -0,0 +1,42 @@
...
+from ..utils import BaseSQLDB
+from .schema import PilotAgents, PilotAgentsDBBase
+
+
+class PilotAgentsDB(BaseSQLDB):
+    metadata = PilotAgentsDBBase.metadata
+
+    async def addPilotReferences(
+        self,
+        pilotRef: list[str],
```
aldbr, line 16: We want to use snake case in diracx: 
```suggestion
        pilot_ref: list[str],
```

diracx-db/src/diracx/db/sql/pilot_agents/db.py
```diff
@@ -0,0 +1,42 @@
...
+        VO: str,
+        gridType: str = "DIRAC",
+        pilotStampDict: dict | None = None,
```
aldbr, line 19:
```suggestion
        grid_type: str = "DIRAC",
        pilot_stamps: dict | None = None,
```

diracx-db/src/diracx/db/sql/pilot_agents/db.py
```diff
@@ -0,0 +1,42 @@
...
+                PilotJobReference=ref,
+                VO=VO,
+                GridType=gridType,
+                SubmissionTime=now,
+                LastUpdateTime=now,
+                Status="Submitted",
+                PilotStamp=stamp,
+            )
+            result = await self.conn.execute(stmt)
+            row_ids.append(result.lastrowid)
```
aldbr, line 40: `lastrowid` is apparently not the preferred way to get the primary key (https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.CursorResult.lastrowid):
- `inserted_primary_key`: https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.CursorResult.inserted_primary_key
- or `returning()`: https://docs.sqlalchemy.org/en/20/core/dml.html#sqlalchemy.sql.expression.Insert.returning

aldbr (commented):
Almost fine, I am just curious about `executemany` versus `execute` n times.

diracx-db/src/diracx/db/sql/pilot_agents/db.py
martynia: I tried this. There are 2 ways of inserting multiple records:
```python
stmt = insert(PilotAgents).values(values)
await self.conn.execute(stmt)
```

and

```python
stmt = insert(PilotAgents)
await self.conn.execute(stmt, values)
```

The first compiles the statement and in a case of an incorrect column name (here `Statuss`) it fails with:
`FAILED diracx-db/tests/pilot_agents/test_pilotAgentsDB.py::test_insert_and_select - sqlalchemy.exc.CompileError: Unconsumed column names: Statuss`

However the second one silently passes. In our case, since the `Status` column has a default value `"Unknown"` we would have an incorrect record.

One can obviously verify the records in the tests, but it would be useful, if the engine does that, no ?

diracx-db/src/diracx/db/sql/pilot_agents/db.py
martynia: We don't need the return value, and if we choose a bulk insert it would not be possible anyway.

diracx-db/src/diracx/db/sql/pilot_agents/db.py
aldbr: Yes I agree!

diracx-db/src/diracx/db/sql/pilot_agents/db.py
```diff
@@ -0,0 +1,44 @@
...
+from sqlalchemy import insert
+
+from ..utils import BaseSQLDB
+from .schema import PilotAgents, PilotAgentsDBBase
+
+
+class PilotAgentsDB(BaseSQLDB):
+    metadata = PilotAgentsDBBase.metadata
+
+    async def addPilotReferences(
```
aldbr, line 14: I forgot this one, sorry: 

```suggestion
    async def add_pilot_references(
```

diracx-db/src/diracx/db/sql/pilot_agents/db.py
aldbr: So I made a quick comparison test and got the following results (I repeated the script multiple times to check that results were consistent):

- `executeasmany`
```
nb_insert,duration
1,0.020946025848388672
2,0.018053054809570312
5,0.025712966918945312
10,0.01901721954345703
20,0.028232574462890625
50,0.03838825225830078
100,0.06927895545959473
200,0.13786983489990234
500,0.2925376892089844
1000,0.6183321475982666
2000,0.9543678760528564
5000,2.5484354496002197
```

- `execute` in a for loop:
```
nb_insert,duration
1,0.014206647872924805
2,0.020319223403930664
5,0.04846644401550293
10,0.03269195556640625
20,0.050367116928100586
50,0.09167718887329102
100,0.1637129783630371
200,0.27317214012145996
500,0.6035165786743164
1000,1.2095882892608643
2000,2.424307346343994
5000,6.025989294052124
```

So your code is fine.

If you want more details about how I ran that test:
- I prepared the following script:

```python
from __future__ import annotations

import asyncio
import time

from diracx.db.sql.pilot_agents.db import PilotAgentsDB


async def setup_pilot_agents_db() -> PilotAgentsDB:
    # Create the PilotAgentsDB instance with a direct MySQL connection
    agents_db = PilotAgentsDB("<url>")

    # Enter the engine context to set up the database connection
    async with agents_db.engine_context():
        async with agents_db.engine.begin() as conn:
            # Create all tables (if necessary)
            await conn.run_sync(agents_db.metadata.create_all)

        # Return the initialized database object (outside the engine context)
        return agents_db

async def test_insert_and_select(pilot_agents_db: PilotAgentsDB):
    # You need to enter the engine_context here again
    for n in (1,2,5,10,20,50,100,200,500,1000,2000,5000):
        start = time.time()

        async with pilot_agents_db.engine_context():
            async with pilot_agents_db as db_session:
                # Add a pilot reference
                refs = [f"ref_{i}" for i in range(n)]
                stamps = [f"stamp_{i}" for i in range(n)]
                stamp_dict = dict(zip(refs, stamps))

                # Add the pilot references with stamps
                await db_session.add_pilot_references(
                    refs, "test_vo", grid_type="DIRAC", pilot_stamps=stamp_dict
                )

                # Add the pilot references without stamps
                await db_session.add_pilot_references(
                    refs, "test_vo", grid_type="DIRAC", pilot_stamps=None
                )
        end = time.time()
        print(f"{n},{end - start}")

async def main():
    # Setup the database
    pilot_agents_db = await setup_pilot_agents_db()

    # Run the test
    await test_insert_and_select(pilot_agents_db)

if __name__ == "__main__":
    # Run the main function using asyncio
    asyncio.run(main())
```

- I ran `run_demo.sh` with your branch
- Once ready, I entered in the `diracx` pod and get the DB URL:

```bash
kubectl get pods # to get the pod id
kubectl exec -it <diracx pod id> bash
$ env | grep PILOT # to get the DB URL
```

- In the script, I replaced `<url>` with the DB URL you have, and from the pod, run the script
- Then I manually changed the code in `db.py` to add the loop again and reran the test

aldbr (approved):

PR #293, opened 2024-09-12
fix: remove SiteMask and SiteMaskLogging

fstagni:
Follows https://github.com/DIRACGrid/DIRAC/pull/7762


fstagni:
Ready for merge, please review (really simple).

PR #296, opened 2024-09-16
Refactor Job manager routes

### LINKED ISSUE #22: JobManagerHandler conversion

ryuwd:
- Renamed from `job_manager` to `jobs`
- Split into multiple modules 
- Removed some endpoints in favour of expressing the desired operation(s) as bulk job status changes
- Major refactor of `reschedule_job` to respect the job state machine
- significantly reduce the number of statements/queries generated in bulk operations

For #22 
Needs https://github.com/DIRACGrid/DIRAC/pull/7794 

closes #22, #341
TODO:
- [x] Tests passing
- [x] Check rescheduling and resetJob
- [x] fix bug in job state machine when rescheduling in JobDB
- [x] make rescheduling and reset job tests a bit more rigorous to check I did not break anything
- [x] regenerate client methods with autorest
- [x] use `config` object to pull out `MaxRescheduling` from the CS instead of making a property in JobDB


diracx-db/src/diracx/db/sql/utils/job_status.py
```diff
@@ -16,11 +16,47 @@
...
+                    job_logging_db,
+                    task_queue_db,
+                    background_task,
+                    force=force,
+                )
+            )
+            for job_id, status_dict in job_update.items()
+        ]
+
+    return {job_id: status for job_id, status in zip(job_update.keys(), results)}
```
chrisburr, line 49: I think you have to call `.result()` on the tasks?

I think this would be clearer:

```suggestion
    async with ForgivingTaskGroup() as tg:
        tasks = {
            job_id: tg.create_task(
                set_job_status(
                    job_id,
                    status_dict,
                    config,
                    job_db,
                    job_logging_db,
                    task_queue_db,
                    background_task,
                    force=force,
                )
            )
            for job_id, status_dict in job_update.items()
        }

    return {k: v.result() for k, v in tasks.items()}
```

ryuwd:
[65da423](https://github.com/DIRACGrid/diracx/pull/296/commits/65da42324681ace611a2618fca1e62a422064bf3) is the start of using bulk insert/update statements in the job manager, and are supposed to be significantly faster than lots of single insert/update statements. I started with rescheduling and job status updates.

diracx-db/src/diracx/db/sql/utils/job_status.py
```diff
@@ -108,112 +119,110 @@ async def reschedule_job(
...
-
-    # Replace the JobID placeholder if any
-    jobJDL = jobJDL.replace("%j", str(job_id))
+    # BULK JDL UPDATE
+    # DATABASE OPERATION
 
     # Update JDL (Should we be doing this here?)
-    result = await job_db.setJobJDL(job_id, jobJDL)
+    # DATABASE OPERATION
+    await job_db.setJobJDLsBulk(jdl_changes)
```
ryuwd, line 212: I should check which rows were successfully updated in the previous operation and exclude unsuccessful updates from the JDL update statement.

ryuwd:
@fstagni do you have any view on things to watch out for with bulk inserts and updates, based on previous experience in DIRAC? I can foresee some risks of things becoming inconsistent, which could be mitigated by reducing the batch sizes or carefully checking the input and results of each statement execution.

diracx-core/src/diracx/core/config/schema.py
```diff
@@ -159,6 +159,56 @@ class OperationsConfig(BaseModel):
...
+    Services: dict[str, Any] | None = None
+    Agents: dict[str, Any] | None = None
+    JobWrapper: dict[str, Any] | None = None
+    Databases: DatabasesConfig
+    Executors: dict[str, Any] | None = None
+    FailoverURLs: dict[str, Any] | None = None
+
+
+class WorkloadManagementConfig(BaseModel):
+    Production: ProductionConfig | None = None
```
aldbr, line 193: I think you can delete this part (since https://github.com/DIRACGrid/DIRAC/pull/7186  was merged)
```suggestion
class WorkloadManagementConfig(BaseModel):
    Production: ProductionConfig | None = None
```

diracx-core/src/diracx/core/config/schema.py
```diff
@@ -159,6 +159,56 @@ class OperationsConfig(BaseModel):
...
+    Configuration: dict[str, Any] | None = None
+    Framework: dict[str, Any] | None = None
+    DataMangement: dict[str, Any] | None = None
+    Monitoring: dict[str, Any] | None = None
+    ProductionManagement: dict[str, Any] | None = None
+    RequestManagement: dict[str, Any] | None = None
+    ResourceStatus: dict[str, Any] | None = None
+    StorageManagement: dict[str, Any] | None = None
+    Transformation: dict[str, Any] | None = None
+    WorkloadManagement: WorkloadManagementConfig | None = None
```
aldbr, line 208: And therefore change this line:

```suggestion
    WorkloadManagement: ProductionConfig | None = None
```

May be `ProductionConfig` is not the most appropriate name here now though. What about having a generic `SystemConfig` with `URLS, Services, Agents, Databases, Executors, FailoverURLs` and a `WorkloadManagementConfig` inheriting from `SystemConfig` with `JobWrapper`?

diracx-core/src/diracx/core/exceptions.py
```diff
@@ -39,6 +39,15 @@ class InvalidQueryError(DiracError):
...
 
 class JobNotFound(Exception):
-    def __init__(self, job_id: int):
+    def __init__(self, job_id: int, detail: str | None = None):
         self.job_id: int = job_id
-        super().__init__(f"Job {job_id} not found")
+        super().__init__(f"Job {job_id} not found" + (" ({detail})" if detail else ""))
+
+
+class JobException(Exception):
```
aldbr, line 47: Based on https://github.com/DIRACGrid/diracx/pull/334, not merged yet:

```suggestion
class JobError(Exception):
```

diracx-testing/src/diracx/testing/__init__.py
```diff
@@ -451,6 +451,50 @@ def with_config_repo(tmp_path_factory):
                 }
             },
             "Operations": {"Defaults": {}},
+            "Systems": {
+                "WorkloadManagement": {
+                    "Production": {
```
aldbr, line 456:
```suggestion
```

diracx-routers/src/diracx/routers/jobs/submission.py
```diff
@@ -0,0 +1,204 @@
...
+            self.userProperties = user_info.properties
+            self.jobDB = None
+            self.allInfo = allInfo
+            self._permissions: dict[str, bool] = {}
+            self._getUserJobPolicy()
+
+    # Check job submission permission
+    policyDict = returnValueOrRaise(DiracxJobPolicy(user_info).getJobPolicy())
+    if not policyDict[RIGHT_SUBMIT]:
+        raise HTTPException(HTTPStatus.FORBIDDEN, "You are not allowed to submit jobs")
```
aldbr, line 103: Is it still needed now that we have the `check_permissions` function at line 80?

diracx-routers/src/diracx/routers/jobs/submission.py
```diff
@@ -0,0 +1,204 @@
...
+    JobDB,
+    JobLoggingDB,
+)
+from ..fastapi_classes import DiracxRouter
+from ..utils.users import AuthorizedUserInfo, verify_dirac_access_token
+from .access_policies import ActionType, CheckWMSPolicyCallable
+
+logger = logging.getLogger(__name__)
+
+router = DiracxRouter(dependencies=[has_properties(NORMAL_USER | JOB_ADMINISTRATOR)])
```
aldbr, line 29: Same here, do we still have to include the `dependencies` parameter here now that we have the access policies?

```suggestion
router = DiracxRouter()
```

diracx-testing/src/diracx/testing/__init__.py
ryuwd: do I remove the entire key?

diracx-routers/src/diracx/routers/jobs/submission.py
ryuwd: Assuming `check_permissions` should do the right thing I removed this

diracx-testing/src/diracx/testing/__init__.py
ryuwd: Where will `MaxRescheduling` be now, do you think?

diracx-core/src/diracx/core/exceptions.py
ryuwd: done!

diracx-routers/src/diracx/routers/jobs/submission.py
ryuwd: hmm, but now we get a failing test:

```
FAILED diracx-routers/tests/test_job_manager.py::test_user_without_the_normal_user_property_cannot_submit_job - AssertionError: [{'JobID': 1, 'MinorStatus': 'Job accepted', 'Status': 'Received', 'TimeStamp': '2024-12-14T19:45:24.667029Z'}]
assert 200 == <HTTPStatus.FORBIDDEN: 403>
 +  where 200 = <Response [200 OK]>.status_code
 +  and   <HTTPStatus.FORBIDDEN: 403> = HTTPStatus.FORBIDDEN
```

diracx-db/src/diracx/db/sql/job/db.py
```diff
@@ -42,6 +47,29 @@ def _get_columns(table, parameters):
...
+        return list(range(start_id, start_id + len(rows)))
+    elif conn.engine.name == "sqlite":
+        # Bulk insert for SQLite
+        if conn.engine.dialect.server_version_info >= (3, 35, 0):
+            results = await conn.execute(table.insert().returning(table.c.JobID), rows)
+            return [row[0] for row in results]
+        else:
+            await conn.execute(table.insert(), rows)
+            start_id = await conn.scalar("SELECT last_insert_rowid()")
+            return list(range(start_id, start_id + len(rows)))
```
ryuwd, line 68: We can't do this!

diracx-core/src/diracx/core/config/schema.py
```diff
@@ -159,6 +159,56 @@ class OperationsConfig(BaseModel):
...
 
 
+class DBConfig(BaseModel):
+    DBName: str
+    Host: str
+    Port: int
+
+
+class JobDBConfig(DBConfig):
+    MaxRescheduling: int = 3
```
ryuwd, line 169: Move this to Services  `JobSchedulingConfig` further up

diracx-db/src/diracx/db/sql/job/db.py
```diff
@@ -189,243 +238,178 @@ async def getJobJDL(self, job_id: int, original: bool = False) -> str:
...
+            JobJDLs.__table__,
+            [
+                {
+                    "JDL": "",
+                    "JobRequirements": "",
+                    "OriginalJDL": compressJDL(original_jdl),
+                }
+                for original_jdl, _ in original_jdls
+            ],
+        )
```
ryuwd, line 298: We need to do single row inserts for this to be safe when creating job IDs. use asyncio taskgroup to do it

diracx-core/src/diracx/core/config/schema.py
ryuwd: And update the tests https://github.com/DIRACGrid/diracx/blob/main/diracx-cli/tests/legacy/cs_sync/integration_test.cfg nd the yaml.

ryuwd:
@chrisburr @aldbr once the tests fully pass I think this is ready to merge

extensions/gubbins/gubbins-db/src/gubbins/db/sql/jobs/db.py
```diff
@@ -40,7 +40,7 @@ async def getJobJDL(  # type: ignore[override]
         info = (await self.conn.execute(stmt)).scalar_one()
         return {"JDL": jdl, "Info": info}
 
-    async def setJobAttributes(self, job_id, jobData):
+    async def setJobAttributesBulk(self, jobData):
```
aldbr, line 43: FYI, this is not yet enforced, but once https://github.com/DIRACGrid/diracx/pull/334 will be merged, we will have to use snake case for function/method names.

diracx-testing/src/diracx/testing/__init__.py
aldbr: > do I remove the entire key?

The key `Production` should be removed, and its value should be associated directly to the `WorkloadManagement` key

extensions/gubbins/gubbins-db/src/gubbins/db/sql/jobs/db.py
ryuwd: (I would like to merge this before it gets too out of sync again and we fall into a cycle of never merging it)

diracx-routers/tests/test_job_manager.py
aldbr, line 1: This file is becoming large and will keep growing through time.
I would split and organize it into different test files the same way you split and organize the logic of the job manager to keep things readable. What do you think?

ryuwd: Very happy to do that - but this PR is getting also very large and unwieldy. I would be happy to defer the refactoring of the tests to an issue / downstream development, but I am afraid I am running out of time to work on this and we run the risk of letting this go on the backburner and going out of sync again. I can make an issue about it

diracx-routers/src/diracx/routers/jobs/status.py
```diff
@@ -0,0 +1,144 @@
...
+    WARNING: This endpoint has been implemented for the compatibility with the legacy DIRAC WMS
+    and the JobCleaningAgent. However, once this agent is ported to diracx, this endpoint should
+    be removed, and a status change to Deleted (PATCH /jobs/status) should be used instead for any other purpose.
+    """
+    await check_permissions(action=ActionType.MANAGE, job_db=job_db, job_ids=job_ids)
+    # TODO: Remove once legacy DIRAC no longer needs this
+
+    # TODO: implement job policy
+    # Some tests have already been written in the test_job_manager,
+    # but they need to be uncommented and are not complete
```
aldbr, line 57: Are these comments still valid?

diracx-client/src/diracx/client/generated/_client.py
```diff
@@ -19,6 +19,7 @@
     AuthOperations,
     ConfigOperations,
     JobsOperations,
+    LollygagOperations,
```
chrisburr, line 22: This looks wrong

extensions/gubbins/gubbins-db/src/gubbins/db/sql/jobs/db.py
aldbr: Yes sure, it just to inform you, but yes I will adjust https://github.com/DIRACGrid/diracx/pull/334 once this PR is merged.

diracx-routers/tests/test_job_manager.py
aldbr: Alright: https://github.com/DIRACGrid/diracx/issues/345

ryuwd: Thanks for that!

diracx-db/src/diracx/db/sql/job/db.py
```diff
@@ -132,7 +124,7 @@ async def setJobAttributes(self, job_id, jobData):
         stmt = update(Jobs).where(Jobs.JobID == job_id).values(jobData)
         await self.conn.execute(stmt)
 
-    async def _checkAndPrepareJob(
```
aldbr, line 135: Do we need this method to be public?

ryuwd: It is used here

https://github.com/DIRACGrid/diracx/pull/296/files/1fb0f3aca4363c53eb1184880710f68baca40fba#diff-fdc6b06128306e37f88bb6e97b63e6bd65f3eec257ebce11d5dc33e6e3511c1eR140

Rescheduling could probably be further refactored, but it can't be moved inside jobdb because of other DB dependencies involved. Something that might be worth revisiting.

diracx-db/src/diracx/db/sql/job/db.py
```diff
@@ -189,243 +206,187 @@ async def getJobJDL(self, job_id: int, original: bool = False) -> str:
...
-                    "MinorStatus": JobMinorStatus.MAX_RESCHEDULING,
-                },
-            )
-            raise ValueError(
-                f"Maximum number of reschedulings is reached: {self.maxRescheduling}"
+                if class_ad_job.lookupAttribute("InputData"):
+                    inputData = class_ad_job.getListFromExpression("InputData")
+                    inputdata_to_insert += [
+                        {"JobID": job_id, "LFN": lfn} for lfn in inputData if lfn
+                    ]
```
aldbr, line 340: I am wondering whether it makes sense to have all this logic in the data access layer rather than in the router (at least part of it).

I haven't checked carefully, but it looks like the part dealing with `JobManifest` and `ClassAd` have not much to do with the DB, am I wrong?

diracx-db/src/diracx/db/sql/job/db.py
aldbr: Ah yes indeed sorry, I did not check that file yet

diracx-db/src/diracx/db/sql/utils/job_status.py
aldbr, line 1: Similar to my previous comment:: https://github.com/DIRACGrid/diracx/pull/296#discussion_r1886986003

I wonder if it makes sense to have such a complex logic on the data access layer side. I would keep the DAL simple (only dealing with DB interactions) and move the complex logic to the router (business logic layer).

It would isolate the responsibilities of each layer
Any opinion?

diracx-db/src/diracx/db/sql/job/db.py
ryuwd: I think that sounds very sensible to me - are you happy for me to move all of this into the router, or perhaps into a utility module in the routers?

diracx-db/src/diracx/db/sql/utils/job_status.py
ryuwd: Same here -- very sensible suggestions. I can move these functions into a routers utility module, if that sounds reasonable?

aldbr: I would "simply" move the content of  `reschedule_jobs_bulk`, `set_job_status_bulk`  and `remove_jobs` into their routes (`routers/jobs/status`): `reschedule_bulk_jobs`, `set_job_statuses` and `remove_bulk_jobs` respectively.

diracx-db/src/diracx/db/sql/job/db.py
aldbr: IIUC:

- This bloc of code could move to the router. Then you would probably call a DB method to insert the original JDLs https://github.com/DIRACGrid/diracx/blob/1fb0f3aca4363c53eb1184880710f68baca40fba/diracx-db/src/diracx/db/sql/job/db.py#L237-L252
- Then the following bloc of code could also go into the router, and may become a utility function such as `check_and_prepare_jdl` (you would remove the `checkAndPrepareJob` function from the `db` class and would add its content into the new `check_and_prepare_jdl`): https://github.com/DIRACGrid/diracx/blob/1fb0f3aca4363c53eb1184880710f68baca40fba/diracx-db/src/diracx/db/sql/job/db.py#L270-L340
- The remaining lines are responsible for inserting the JDL, the job and its inputs into the DB, and should become the content of `insert_bulk` I would say.

What do you think?

ryuwd: Done, but JobDB tests relied on most or all of this logic being inside job DB... can't import the utility fn from routers because of (presumably) a cyclic dependency. Even more confusing as everything runs just fine locally for me. It is very tempting to `pytest.skip`... but it could be something we could resolve by tweaking the CI environment too. what do now?

ryuwd: We agreed to keep these as utility functions inside `diracx-db` in case agents will need to call them (i.e. not via the API/services)

aldbr (approved):

PR #299, opened 2024-09-17
feat: lock file while read, write and refresh token

### LINKED ISSUE #133: Use file locking to prevent multiple processes using the refresh token at the same time

natthan-pigoux:
closes #133

Tests can be performed on `read/write_credentials`, and the client (probably simpler if we do it through `get_token` directly). Test cases:
- [x] (`read/write_credentials` and `get_token`) multiple processes attempting to read or write to the credentials file simultaneously.
- [x] (`read/write_credentials`) credentials file is missing, empty, invalid.
- [x] (`get_token`) token exists and is valid, credentials file is missing, empty, invalid, token needs to be refreshed, refresh token valid/invalid.

### DIFF
--- a/diracx-cli/src/diracx/cli/__init__.py
+++ b/diracx-cli/src/diracx/cli/__init__.py
@@ -10,7 +10,7 @@ from diracx.client.aio import DiracClient
 from diracx.client.models import DeviceFlowErrorResponse
 from diracx.core.extensions import select_from_extension
 from diracx.core.preferences import get_diracx_preferences
-from diracx.core.utils import write_credentials
+from diracx.core.utils import read_credentials, write_credentials
 
 from .utils import AsyncTyper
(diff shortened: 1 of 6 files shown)


aldbr:
@natthan-pigoux 

- I fixed the `mypy` issues
- I split the test in 2 different files to avoid importing `diracx-client` within `diracx-core`
- I simplified a few parts in the tests (by using the `tmp_path` fixture for instance)
- The logic to test file locking looked robust in terms of evolutivity but sounds complex to debug and flaky (e.g. use of `time.sleep`). I decided to mock `flock` instead, it's less robust (we would need change the tests if we use something else), but it looks like more deterministic, quicker, and easier to understand. If you are against that change, we can discuss.

chaen (approved):

PR #312, opened 2024-10-17
feat(auth): add scope documentation

aldbr:
Add documentation about the scope following https://github.com/DIRACGrid/diracx/pull/300

The same string is repeated at 3 different locations in the code. While it is cumbersome (developers would need to update every instances of the doc to avoid any discrepancy, which will likely happen at some point), I don't see any other way of doing that as openapi reads the static documentation and I find it convenient to have the details when using the CLI or the openapi document.

Any opinion?

### DIFF
--- a/diracx-cli/src/diracx/cli/__init__.py
+++ b/diracx-cli/src/diracx/cli/__init__.py
@@ -38,12 +38,34 @@ def vo_callback(vo: str | None) -> str:
 
 @app.async_command()
 async def login(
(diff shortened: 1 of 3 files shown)


chaen:
I think it is fine as it is. If really  it is too much of an issue, you can put this in a constant, and set the `__docs__` of a function to that constant. But I really don't think it's needed

chaen (approved):

Issue #34, opened 2023-06-30
Enable more ruff rulesets


chrisburr:
Currently we're missing a bunch of rules to ruff which are probably useful, e.g. `RUF`.

PR #334, opened 2024-12-02
feat(pyproject): add `N` (pep8 naming) rule to ruff config...

aldbr:
...Before it becomes too inconsistent.

Partially solves https://github.com/DIRACGrid/diracx/issues/34

It seems there is no clear choice for the attributes of the SQL Alchemy classes and we actually write them in 2 different ways:

- snake case (this is how it is presented in the SQLAlchemy documentation): https://github.com/DIRACGrid/diracx/blob/34095d8a37f5a9005a0381df054fa5f6fd4d8bae/diracx-db/src/diracx/db/sql/auth/schema.py#L40-L48

- camel case: https://github.com/DIRACGrid/diracx/blob/34095d8a37f5a9005a0381df054fa5f6fd4d8bae/diracx-db/src/diracx/db/sql/job/schema.py#L65-L70

We should probably make an opinionated choice and specify it in `CODING_CONVENTION.md`.
Any opinion?

Needs: https://github.com/DIRACGrid/DIRAC/pull/7968


fstagni:
snake_case ...?

chaen:
`snake_case` definitely, BUT the reason for having the Camel case is the compatibility with DIRAC :-) it just started because of the prototype.
So indeed, the attributes should be xnake case, but the column name has to be Caml case

aldbr:
Alright, I converted the attributes to snake_case and the column name in PascalCase.
So even the `Auth` SQLAlchemy classes, which are new, follow the same pattern in order to remain consistent.

aldbr:
I guess this error is expected: https://github.com/DIRACGrid/diracx/actions/runs/12135284425/job/33834028304?pr=334#step:14:205

Because I modified the DB schema of `AuthDB`. The pipeline should be fixed once the PR will be merged if I understand correctly (the new schema will be included in the `diracx/services:dev` image).

chrisburr (approved):

PR #356, opened 2024-12-19
Fix job attribute update to account for mismatching columns between rows

ryuwd:
cc @aldbr


ryuwd:
OK, test added @aldbr !

aldbr (approved):

PR #364, opened 2025-01-10
fix: added jsonrpc for autorest

fstagni:


fstagni:
The CI for the other PRs is failing to run `autorest`, with message like

```
 Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/runner/.autorest/@autorest_python@6.27.3/node_modules/@autorest/python/autorest/jsonrpc/server.py", line 12, in <module>
    from jsonrpc import dispatcher, JSONRPCResponseManager
ModuleNotFoundError: No module named 'jsonrpc'
```

so I simply added `jsonrpc` in this PR and now autorest works. It turns out that now `pip check` fails. So, question is, should I add `jsonrpc` in the container images?

fstagni:
> The CI for the other PRs is failing to run `autorest`, with message like
> 
> ```
>  Traceback (most recent call last):
>   File "<frozen runpy>", line 198, in _run_module_as_main
>   File "<frozen runpy>", line 88, in _run_code
>   File "/home/runner/.autorest/@autorest_python@6.27.3/node_modules/@autorest/python/autorest/jsonrpc/server.py", line 12, in <module>
>     from jsonrpc import dispatcher, JSONRPCResponseManager
> ModuleNotFoundError: No module named 'jsonrpc'
> ```
> 
> so I simply added `jsonrpc` in this PR and now autorest works. It turns out that now `pip check` fails. So, question is, should I add `jsonrpc` in the container images?

Oh and I did not check why it "suddenly" fails, it's probably because of some dependency dropped somewhere.

fstagni:
Looks like not needed.

PR #397, opened 2025-02-05
feat: patch job metadata endpoint

### LINKED ISSUE #162: Add routes for JobStateUpdateHandler

ryuwd:
Todo:
- [x] Write `PATCH /metadata`
- [x] Tests
  - [x] test patching attributes and parameters
  - [x] test getting job status history in `/search`

Closes #162

### DIFF
--- a/diracx-logic/src/diracx/logic/jobs/query.py
+++ b/diracx-logic/src/diracx/logic/jobs/query.py
@@ -40,7 +40,10 @@ async def search(
     if query_logging_info := ("LoggingInfo" in (body.parameters or [])):
         if body.parameters:
             body.parameters.remove("LoggingInfo")
-        body.parameters = ["JobID"] + (body.parameters or [])
+            if not body.parameters:
+                body.parameters = None
+            else:
+                body.parameters = ["JobID"] + (body.parameters or [])
 
     # TODO: Apply all the job policy stuff properly using user_info
     if not config.Operations["Defaults"].Services.JobMonitoring.GlobalJobsInfo:
--- a/diracx-logic/src/diracx/logic/jobs/status.py
+++ b/diracx-logic/src/diracx/logic/jobs/status.py
@@ -486,7 +486,7 @@ async def set_job_parameters_or_attributes(
 ):
     """Set job parameters or attributes for a list of jobs."""
     possible_attribute_columns = [
-        name.lower() for name in _get_columns(Jobs.__table__, None)
+        col.name.lower() for col in _get_columns(Jobs.__table__, None)
     ]
 
     attr_updates = {}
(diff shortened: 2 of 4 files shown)


diracx-routers/src/diracx/routers/jobs/status.py
```diff
@@ -135,3 +138,43 @@ async def reschedule_bulk_jobs(
...
+        # else set elastic parameters DB
+        param_updates[job_id] = {
+            pname: pvalue
+            for pname, pvalue in metadata.items()
+            if pname.lower() not in possible_attribute_columns
+        }
+    # bulk set job attributes
+    await job_db.set_job_attributes_bulk(attr_updates)
+
+    # TODO: can we upsert to multiple documents?
```
chrisburr, line 174: Can you open an issue for this?

diracx-api/src/diracx/api/jobs.py
```diff
@@ -30,8 +30,8 @@ async def create_sandbox(paths: list[Path], *, client: DiracClient) -> str:
     The returned value is the PFN of the sandbox in the storage backend and can
     be used to submit jobs.
     """
-    with tempfile.TemporaryFile(mode="w+b") as tar_fh:
-        with tarfile.open(fileobj=tar_fh, mode=f"w|{SANDBOX_COMPRESSION}") as tf:
+    with tempfile.TemporaryFile(mode="w+b") as tar_fh:  # type: ignore
+        with tarfile.open(fileobj=tar_fh, mode=f"w|{SANDBOX_COMPRESSION}") as tf:  # type: ignore
```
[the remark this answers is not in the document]
chaen: yes, I'll finish my PR tomorrow and then you can just rebase

diracx-routers/src/diracx/routers/jobs/status.py
ryuwd: #401

ryuwd:
Rebased and ready to go

ryuwd:
@aldbr LGTM, please go ahead

diracx-logic/src/diracx/logic/jobs/query.py
```diff
@@ -40,7 +40,10 @@ async def search(
     if query_logging_info := ("LoggingInfo" in (body.parameters or [])):
         if body.parameters:
             body.parameters.remove("LoggingInfo")
-        body.parameters = ["JobID"] + (body.parameters or [])
+            if not body.parameters:
+                body.parameters = None
```
fstagni, line 44: Looks like an impossible thing to happen

diracx-logic/src/diracx/logic/jobs/status.py
```diff
@@ -474,3 +476,41 @@ async def remove_jobs_from_task_queue(
...
+
+async def set_job_parameters_or_attributes(
+    updates: dict[int, dict[str, Any]],
+    job_db: JobDB,
+    job_parameters_db: JobParametersDB,
+):
+    """Set job parameters or attributes for a list of jobs."""
+    possible_attribute_columns = [
+        col.name.lower() for col in _get_columns(Jobs.__table__, None)
+    ]
```
fstagni, line 489: Is this (`lower()` of all column names) and the subsequent check needed? Can't we compare without lowering?

diracx-logic/src/diracx/logic/jobs/query.py
ryuwd: This in essence forces `body.parameters` to `None` if removing `LoggingInfo` leaves an empty list, hence making it as if the default set of parameters (in addition to the logging info) are the ones should be queried when calling `job_db.search`. I ran into an unfortunate edge case in getting the unit tests to pass and this is what I came up with.

ryuwd: The default parameter set includes `JobID` and a bunch of other things, but if you don't keep this then you end up with `body.parameters = ["JobID"]` only

diracx-logic/src/diracx/logic/jobs/status.py
ryuwd: I'm inclined to agree. I just wonder what would happen if somebody passed a case-sensitive un-matching attribute column that then gets registered in the JobParameterDB instead instead of throwing an error. Maybe we should throw an exception if the case-sensitive vs. insensitive matches don't line up

fstagni: I see, nothing is perfect. Do as you feel.

ryuwd: done

ryuwd:
I think this is good to go now. Thanks for the help @aldbr @fstagni !

fstagni (approved):

PR #389, opened 2025-01-30
fix helm values generation if cfg contains NoSetup

natthan-pigoux:
closes #388

PR #398, opened 2025-02-06
fix helm values generation if cfg contains NoSetup

### LINKED ISSUE #388: cli: generate helm values fails if no Setup in cfg
`dirac internal legacy generate-helm-values` fails if there is no "Setup" in the CS file.
@chaen was it wanted?

natthan-pigoux:
Reopenning #389 due to conflict. Closes #388.

### DIFF
--- a/diracx-cli/src/diracx/cli/internal/legacy.py
+++ b/diracx-cli/src/diracx/cli/internal/legacy.py
@@ -240,14 +240,14 @@ def generate_helm_values(
     default_db_password = cfg["Systems"].get("Databases", {}).get("Password")
 
     all_db_configs = {}
-    if cfg["DIRAC"].get("NoSetup") == "True":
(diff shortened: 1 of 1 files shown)


chaen:
Can you please check if you still have the problem with the current `master` branch ? I have the impression it is already fixed

fstagni:
@natthan-pigoux this PR has conflicts right now. At the same time, I do not fully understand what you want to do, because
diracx can only interact with v9 CS, and in there there is no mention of setup (nor `DefaultSetup` as you used here). The `NoSetup` flag is there purely for backward compatibility with a v8 client and there's really no need to bother about it in diracx. So, unless I am missing something, this PR can be closed.

natthan-pigoux:
Hi @fstagni , sorry I completely forgot to answer this PR. Yes it can be closed.

Issue #348, opened 2024-12-17
Moving towards a 3-tier layer architecture?


aldbr:
**Current state**

In the backend, we have 2 layers:
- the routers, presenting the API to the users, and containing the business logic
- the dbs, also containing some business logic and dealing with the dbs through an ORM

**Context:**

For now it's fine, but in the near future we are going to include:
- Agents: as they are hosted in the backend, they could be allowed to bypass routers and directly call the source code to avoid going through the network if it's not necessary
- Similarly, we will have services depending on other services, such as the `production manager` depending on the `transformation manager` depending on the `job manager`. To go faster, they could be allowed to bypass networking and call directly the source code.

**Problem**

In the good old Dirac, it seems like we implicitly made the choice to merge the business logic with the data access layer, which creates tremendous DB interfaces, not always easy to follow. Also, as this is - it seems to me at least - implicit, we sometimes find some logic in both the Services and the DB interfaces (we can even see that in the only service we imported to DiracX).

**Potential solution**

In order to have a clear separation of concerns and a clear organization of the code, given our requirements, I would go towards a new layer, so that we would have:
- Presentation layer: `diracx-routers` and FastAPI, that would be in charge of the API and the presentation of the information returned (e.g. HTTP Exceptions)
- Business Logic layer: `diracx-business` (?) that would simply contain Dirac business logic (potentially allowing us to move from whatever framework we find in the future in an "easier" way). This layer would call the DAL to store information.
- Data Access layer: `diracx-dbs`, which would only contain operations with the DBs and nothing else.

```mermaid
flowchart LR
    dbs["diracx-dbs (sqlalchemy/os)"] -- uses --> core["diracx-core (domain)"]
    business["diracx-business (Dirac)"] -- uses --> dbs & core
    services["diracx-services (FastAPI)"] -- uses --> business & core
    services -- generates --> openapi["OpenAPI"]
    tasks["diracx-tasks (celery?)"] -- uses  --> business & core
    client["diracx-client (autorest)"] -- consumes --> openapi
    client -- uses --> core
    api["diracx-api"] -- uses --> client
    cli["diracx-cli (typer)"] -- uses --> api & client
     dbs:::Pine
     business:::Pine
     services:::Pine
     openapi:::Rose
     tasks:::Pine
     client:::Sky
     api:::Sky
     cli:::Sky
    classDef Rose stroke-width:1px, stroke-dasharray:none, stroke:#FF5978, fill:#FFDFE5, color:#8E2236
[...]

PR #403, opened 2025-02-25
feat: add a business logic layer

aldbr:
Details explained in https://github.com/DIRACGrid/diracx/issues/348

- [x] split `diracx-db` into 2 sub-directory: 
  - `bl` and `dal` (names should probably be `business_logic` and `data_access`).
  - `diracx-routers` becomes the presentation layer: it solely calls the business logic layer and convert the returned results/raised exceptions into HTTP-friendly responses for the end users.
- [x] auth: adapt `diracx-routers` and move business logic to `bl`. 
- [ ] jobs: adapt `diracx-routers` and move business logic to `bl`. 
- [ ] check if we can enforce consistency via programmatic tools:
  - developers should not be allowed to write business logic code in the presentation layer
  - developers should not be allowed to write data access code in the business logic layer
- [ ] documentation about the 3-layer architecture within diracx


aldbr:
Oops, I thought we said we would add a business logic layer within `diracx.db`: https://github.com/DIRACGrid/diracx/issues/348#issuecomment-2550754120

I thought it would be better to be consistent and have all the business logic at the same place to have a single responsibility principle between the layers + follow a standard architectural pattern.

aldbr:
- First of all, sorry this PR is actually massive and very hard to review.

- `Deployment/docker` and `Integration Tests` are not passing, I think they would pass once https://github.com/DIRACGrid/container-images/pull/31 will be merged (should be safe to merge IIUC).

chaen:
You can create the dev tag yourself by triggering the deployment :-) Even if it creates a new real tag, it does not really matter... :-)

aldbr:
Alright, all is :green_circle: now

diracx-core/src/diracx/core/settings.py
```diff
@@ -102,3 +113,75 @@ class DevelopmentSettings(ServiceSettingsBase):
...
+    allowed_redirects: list[str] = []
+    device_flow_expiration_seconds: int = 600
+    authorization_flow_expiration_seconds: int = 300
+
+    # State key is used to encrypt/decrypt the state dict passed to the IAM
+    state_key: FernetKey
+
+    # TODO: this should probably be something mandatory
+    # to set by the user
+    token_issuer: str = "http://lhcbdirac.cern.ch/"  # noqa: S105
```
fstagni, line 139: Is this something used (e.g. by the tests) or something here as an example?

diracx-core/src/diracx/core/models.py
fstagni, line 1: I know you did not start this file, anyway: what is the rationale for having all (?) models in it?

diracx-db/src/diracx/db/sql/sandbox_metadata/db.py
```diff
@@ -50,11 +54,9 @@ def get_pfn(bucket_name: str, user: UserInfo, sandbox_info: SandboxInfo) -> str:
         return "/" + "/".join(parts)
 
     async def insert_sandbox(
-        self, se_name: str, user: UserInfo, pfn: str, size: int
+        self, owner_id: int, se_name: str, pfn: str, size: int
     ) -> None:
         """Add a new sandbox in SandboxMetadataDB."""
-        # TODO: Follow https://github.com/DIRACGrid/diracx/issues/49
```
fstagni, line 56: The linked issue is not solved. It does not apply here anymore?

diracx-db/tests/auth/test_authorization_flow.py
```diff
@@ -58,11 +58,6 @@ async def test_insert_id_token(auth_db: AuthDB):
                 uuid, id_token, MAX_VALIDITY
             )
 
-    # We shouldn't be able to retrieve it twice
-    async with auth_db as auth_db:
-        with pytest.raises(AuthorizationError, match="already used"):
-            res = await auth_db.get_authorization_flow(code, MAX_VALIDITY)
-
```
fstagni, line 65: Why remove this test?

diracx-db/pyproject.toml
```diff
@@ -27,15 +25,15 @@ testing = [
     "diracx-testing",
 ]
 
-[project.entry-points."diracx.db.sql"]
+[project.entry-points."diracx.dbs.sql"]
```
fstagni, line 28: I am confused: is it `db` or `dbs`?

fstagni (commented):
It is close to impossible to properly review this PR, so these are just few comments.

For a PR that mostly should be a lot of "move around" of code, I am surprised that there are more than 700 lines of code added

diracx-core/src/diracx/core/settings.py
aldbr: I don't think it's used, let's try to remove it and see if it breaks

diracx-core/src/diracx/core/models.py
aldbr: Good question actually, I think we should agree on the purpose of this file and it would deserve a docstring.
I added all the models that were (i) passed as inputs to the routers; (ii) returned as outputs by the routers, so the ones that could be shared between the clients (`cli`, `api`) and the services (`routers`, `logic`, `db`).

At some point we should probably make a `models` directory with multiple modules such as `jobs`, `auth`...

diracx-db/src/diracx/db/sql/sandbox_metadata/db.py
aldbr: The linked issue is not solved:
- I break down the method into smaller methods in `diracx-db` to move the logic from it to `diracx-logic`.
- The comment was moved to `diracx-logic` because it makes more sense to read it from here now.

diracx-db/tests/auth/test_authorization_flow.py
aldbr: Because:
- I break down `get_authorization_flow` into smaller methods to move the logic from `diracx-db` to `diracx-logic`
- The new `get_authorization_flow` is now very simple and does not require complex testing as it was the case before: complex testing should now be done in `diracx-routers` and it's already done.

diracx-db/pyproject.toml
aldbr: It's whatever we want as long as we are consistent.

We have:
- `project.entry-points."diracx.services"`
- `project.entry-points."diracx.access_policies"`

So I thought it would be more consistent to have every entry points using plural.

chrisburr (approved):

PR #406, opened 2025-03-06
chore: remove __all__ from non- __init__ files

fstagni:
I do not see why they would be useful, but convince me otherwise.

### DIFF
--- a/diracx-api/src/diracx/api/jobs.py
+++ b/diracx-api/src/diracx/api/jobs.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-__all__ = ("create_sandbox", "download_sandbox")
-
 import hashlib
 import logging
 import os
--- a/diracx-api/src/diracx/api/utils.py
+++ b/diracx-api/src/diracx/api/utils.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-__all__ = ("with_client",)
-
 from functools import wraps
 
 from diracx.client.aio import DiracClient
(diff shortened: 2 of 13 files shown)


fstagni:
> `__all__` lets us define the public API of each module.

It is only used when you do `from module import *`

chrisburr:
It's also used by some tooling, for example pylance for editor autocomplete: https://github.com/microsoft/pylance-release/blob/392d64e0029c7e0adbb8b6baa2b6a0ce3e3a4c90/docs/settings/python_analysis_packageIndexDepths.md#how-to-use-pythonanalysispackageindexdepths

fstagni:
IMHO, that's not a particularly strong point. 
I see few pros and some cons, mostly the added boilerplate. It's of very little use if we do not use `import *`, which we don't because is generally discouraged.

chaen:
It's a way of making the public API explicit, and as @chrisburr said, it is used by autocomplete. 
I am not in favor of forcing it to be there, but if it is there, I would not remove it

aldbr:
@fstagni created this PR because he noticed that `JobParametersDB` was not listed in `__all__` within `diracx.routers.dependencies`. It means that we are forgetting to add new variables/functions in the public API over time, which might possibly create inconsistencies.

- If there is an automatic rule (`mypy`?) checking that imported variables and functions are all part of `__all__` if it exists, then we should probably go for it (though I haven't found anything by doing a quick search)
- Else we should at least provide a piece of documentation about that for the developers (`CODING_CONVENTION.md`)

PR #408, opened 2025-03-10
fix: remove DIRAC setup

fstagni:
(also using this to test failures in integration tests)

### DIFF
--- a/diracx-cli/tests/legacy/cs_sync/integration_test.cfg
+++ b/diracx-cli/tests/legacy/cs_sync/integration_test.cfg
@@ -9,21 +9,18 @@ DIRAC
   }
   Setups
   {
-    dirac-JenkinsSetup
-    {
-      Accounting = Production
-      Configuration = Production
-      DataManagement = Production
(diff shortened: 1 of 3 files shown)


fstagni:
Here there are test fixes. I merge already.

diracx-cli/tests/legacy/cs_sync/integration_test.cfg
```diff
@@ -9,21 +9,18 @@ DIRAC
...
-      StorageManagement = Production
-      Production = Production
-      Transformation = Production
-      WorkloadManagement = Production
-      Tornado = Production
-    }
+    Accounting = Production
+    Configuration = Production
+    DataManagement = Production
+    Framework = Production
```
chaen, line 15: But why is this still needed ? There is no setup anymore, so we don't need this `<System> = Production` anymore, do we ?

fstagni: Ah, good point, this can be removed indeed.

PR #410, opened 2025-03-12
fix: Moving the coding convention documentation and fixing the db location

Robin-Van-de-Merghel:
- Moving the part from `/docs/README.md` to `/docs/CODING_CONVENTION.md`
- Coding convention documentation restructured : from sections to a table
- Adding clear text around the instructions
- In `/docs/README.md`, changing the database classes path

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
       - id: check-added-large-files
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.9.10'
+    rev: 'v0.11.0'
     hooks:
       - id: ruff
         args: ["--fix"]


docs/README.md
```diff
@@ -87,9 +87,10 @@ This only runs the diracx server, not any dependency like external IdP.
 
 ## Add a DB
 
-Database classes live in `src/diracx/db/<dbname>`. Have a look at the `src/diracx/db/dummy/` to see how to implement your own DB
+Database classes live in `src/diracx/db/sql/<dbname>`. Have a look at the `src/diracx/db/sql/dummy/` to see how to implement your own DB.
```
aldbr, line 90: You also need to include opensearch DBs (e.g. https://github.com/DIRACGrid/diracx/blob/main/diracx-db/src/diracx/db/os/job_parameters.py)

What about: 
```suggestion
Database classes live in `src/diracx/db/<dbtype>/<dbname>`. Have a look at the `src/diracx/db/sql/dummy/` to see how to implement your own DB.
```

Robin-Van-de-Merghel (commented):
I added more details now.

docs/CODING_CONVENTION.md
```diff
@@ -24,63 +43,105 @@ def my_ficture():
...
+    pass
+
+def block():
+    pass
+
+def aPrivateOrUnusefulFunction():
+    pass
+```
 
+Using `from diracx-router import *` will only import foo and bar, but not baz.
```
fstagni, line 134:
```suggestion
Using `from diracx-router import *` will only import `redirect` and `block`, but not aPrivateOrUnusefulFunction`.
```

PR #423, opened 2025-03-27
feat: Adding commit semantic to the workflows

Robin-Van-de-Merghel:
Maybe we can add it to the pre-commit as well, tried it but did not succeeded.


fstagni:
There are more standard ways of doing this, the github marketplace should provide.

chrisburr:
Replaced by #540

PR #429, opened 2025-03-31
Add support for heartbeats and job commands

chrisburr:


diracx-logic/src/diracx/logic/jobs/status.py
```diff
@@ -39,6 +42,15 @@
...
 
+SQL_HEARTBEAT_FIELDS = [
+    "LoadAverage",
+    "MemoryUsed",
+    "Vsize",
+    "AvailableDiskSpace",
+    "CPUConsumed",
+    "WallClockTime",
+]
+
```
ryuwd, line 53: would it be better to pull this out of the Table object instead, or from a list defined in `db.py` that pulls this out from the Table object?

diracx-routers/src/diracx/routers/jobs/access_policies.py
```diff
@@ -45,6 +47,11 @@ async def policy(
         assert action, "action is a mandatory parameter"
         assert job_db, "job_db is a mandatory parameter"
 
+        if action == ActionType.PILOT:
+            # TODO: For now we map this to MANAGE but it should be changed once
```
chaen, line 51: https://github.com/DIRACGrid/diracx/issues/444

PR #431, opened 2025-03-31
fix(diracx-client): create JSON from _patch

aldbr:
The CI is currently failing because the `JSON` type we rely on in `_patch` does not exist anymore.


chaen:
You are just re-running the client generation right ?

chrisburr:
The change was caused by https://github.com/pydantic/pydantic/pull/11392

chrisburr:
For reference this is the change in the `openapi.json`:

```json
diff --git a/tmp/cburr/a.json b/tmp/cburr/b.json
index 0189ca7..f96f562 100644
--- a/tmp/cburr/a.json
+++ b/tmp/cburr/b.json
@@ -917,6 +917,7 @@
               "application/json": {
                 "schema": {
                   "type": "object",
+                  "additionalProperties": true,
                   "title": "Response Jobs Reschedule Jobs"
                 }
               }
@@ -937,6 +938,7 @@
             "application/json": {
               "schema": {
                 "additionalProperties": {
+                  "additionalProperties": true,
                   "type": "object"
                 },
                 "type": "object",
@@ -1064,7 +1066,8 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "type": "object"
+                    "type": "object",
+                    "additionalProperties": true
                   },
                   "title": "Response Jobs Search"
                 },
@@ -1131,7 +1134,8 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "type": "object"
+                    "type": "object",
+                    "additionalProperties": true
                   },
                   "title": "Response 206 Jobs Search"
                 }
@@ -2002,6 +2006,7 @@
             "title": "Dirac Group"
           },
           "policies": {
+            "additionalProperties": true,
             "type": "object",
             "title": "Policies"
           },
```

aldbr:
> You are just re-running the client generation right ?

Yes, but there is a discrepancy between my local environment and the CI for `gubbins`.
The client generation is fine for me locally but fails in the CI.

chaen:
Btw @aldbr this was an overlook. There should not be anything really in the `_patch` file. You need to have the logic in the `patches` directory, and then in the `_patch` import it

chrisburr (approved):

PR #437, opened 2025-04-01
Add DIRACX_SANDBOX_STORE_SE_NAME in the helm value

chaen:


diracx-cli/src/diracx/cli/internal/legacy.py
```diff
@@ -364,6 +364,13 @@ def generate_helm_values(
     diracx_settings["DIRACX_SANDBOX_STORE_BUCKET_NAME"] = (
         f"{diracx_hostname.split('.')[0]}-sandboxes"
     )
+    try:
+        diracx_settings["DIRACX_SANDBOX_STORE_SE_NAME"] = cfg["Systems"][
+            "WorkloadManagement"
+        ]["Services"]["SandboxStore"]["LocalSE"]
```
fstagni, line 370:
```suggestion
        diracx_settings["DIRACX_SANDBOX_STORE_SE_NAME"] = cfg["Systems"][
            "WorkloadManagement"
        ]["Services"]["SandboxStore"].get("LocalSE", "SandboxSE")
```

chaen: That's not needed, as it already is the default in the pydantic model

chrisburr (approved):

PR #439, opened 2025-04-02
fix: Log failed CS git pull attempts

ryuwd:
logs failed `git pull` attempts

use `git clone [--branch ...] repo [location]` instead of `git clone ... && git checkout ...`

also fixed a typo


ryuwd:
I discussed with @chaen and we decided to keep the TTLCache for now

PR #441, opened 2025-04-03
Improve client extensions

chrisburr:
The main thing here is that this modifies the client to be extendable in more ways. In particular it's now possible to:

* Add new routers
* Patch methods in new routers
* Add new methods to existing routers
* Patch new methods in existing routers
* Patch methods which are already patched in DiracX (via inheritance and `super()`)
* Patch the `SyncDiracClient` and `AsyncDiracClient` itself (known as `{As,S}yncExtensionClient` where `Extension` is replaced with the name of the extension).
* Anything which uses the plain DiracX client automatically gets promoted to using the `{As,S}yncExtensionClient` instead making it possible to inject patches into code which only lives in plain DiracX.

To see how this works, see the documentation in `_client_importer.py`.

I propose this is ready now, the remaining parts are minor and can be tracked in: https://github.com/DIRACGrid/diracx/issues/442


diracx-client/pyproject.toml
```diff
@@ -19,21 +19,26 @@ dynamic = ["version"]
...
-
-[build-system]
-requires = ["setuptools>=61", "wheel", "setuptools_scm>=8"]
-build-backend = "setuptools.build_meta"
-
 [project.entry-points."diracx"]
 client_class = "diracx.client.generated._client:Dirac"
 aio_client_class = "diracx.client.generated.aio._client:Dirac"
 
+[build-system]
```
chaen, line 26: Maybe comment here that this is because of the pth file

diracx-client/pyproject.toml
```diff
@@ -19,21 +19,26 @@ dynamic = ["version"]
...
 
-[tool.setuptools.packages.find]
-where = ["src"]
-
-[build-system]
-requires = ["setuptools>=61", "wheel", "setuptools_scm>=8"]
-build-backend = "setuptools.build_meta"
-
 [project.entry-points."diracx"]
 client_class = "diracx.client.generated._client:Dirac"
```
chaen, line 23: This isn't needed anymore is it ?

diracx-client/pyproject.toml
chrisburr: I was going to open an issue to move everything to using `hatchling`, it was less clear two years ago but it's now faster and objectively better than setuptools.

diracx-client/pyproject.toml
chrisburr: Indeed!

extensions/README.md
```diff
@@ -10,7 +10,6 @@ If you write your own extension, just replace `gubbins` with whatever name you c
 The structure of the repo, the content of the `pyproject.toml` files, the `__init__` of the modules... are all skeletons that you must reproduce.
 
 Extensions are controled via:
-* the `DIRACX_EXTENSIONS` environment variable. It should be set to `gubbins,diracx`
 * the entrypoints found in the various `pyproject.toml`
```
chaen, line 13: Did you add somewhere about the new entrypoint ?

extensions/gubbins/gubbins-client/src/gubbins/client/_generated/aio/operations/_patch.py
```diff
@@ -0,0 +1,50 @@
...
+# from diracx.client._generated.aio.operations._patch import *
+# from diracx.client._generated.aio.operations._patch import __all__ as _patch_all
+
+
+class WellKnownOperations(_WellKnownOperations):
+    def my_method(self): ...
+
+
+__all__ = ["WellKnownOperations"]
+# __all__ += [x for x in _patch_all if x not in __all__]
```
chaen, line 27: I don't think you meant to commit that

diracx-testing/src/diracx/testing/client_generation.py
```diff
@@ -0,0 +1,140 @@
...
+    # This is required to be able to work offline
+    # TODO: if offline, find the version already installed
+    # and use it
+    # cmd += [f"--use=@autorest/python@{AUTOREST_VERSION}"]
+
+    # ruff: disable=S603
+    subprocess.run(cmd, check=True)
+
+    if extension_name != "diracx":
+        # For now we don't support extending the models in extensions. To make
```
chaen, line 114: Why not ? We said we would just do a super set of all of them

extensions/gubbins/gubbins-client/src/gubbins/client/_generated/aio/operations/_patch.py
chrisburr: Indeed, fixed (though this will will be autogenerated as part of https://github.com/DIRACGrid/diracx/issues/442).

diracx-testing/src/diracx/testing/client_generation.py
chrisburr: It's included in https://github.com/DIRACGrid/diracx/issues/442 (though I'm actually not convinced it will ever be used...)

extensions/README.md
chrisburr: Done in cdddada

PR #447, opened 2025-04-07
Various JobDB fixes

chrisburr:


diracx-db/src/diracx/db/sql/job/db.py
```diff
@@ -167,6 +167,24 @@ async def update_job_jdls(self, jdls_to_update: dict[int, str]):
             ],
         )
 
+    @staticmethod
+    def _set_job_attributes_fix_value(column, value):
+        """Apply corrections to the values before inserting them into the database.
+
+        TODO: Move this logic into the sqlalchemy model.
```
chrisburr, line 174: https://github.com/DIRACGrid/diracx/issues/448

chaen (approved):

Issue #444, opened 2025-04-04
Update WMSJobPolicy with proper behavior for PILOT action


chaen:

Robin-Van-de-Merghel:
For that, I had a question about policies. Currently, I did not find a way to have multiple policies to avoid code duplication. (Tell me if I did not find a way that existed). But currently we need two policies about pilots:

- Pilots need a policy to see if they are registered, [especially for logging](https://github.com/DIRACGrid/diracx/blob/78e00aa57f4191034dbf643c7ed2857a93b53f60/diracx-routers/src/diracx/routers/pilots/access_policies.py#L28-L86) (in the `/api/pilots` endpoints)
- In `/api/jobs` we will also need to make sure that a pilot is registered, but with another policy that will also support regular Users (`WMSJobPolicy`)

Do we want to add a code duplication in `WMSJobPolicy` for Pilots, or do we prefer using some kind of delegation that we would implement in policies for future updates?

Example: `WMSJobPolicy`: If it's a user, delegation to the user policy, else delegation to the pilot policy

PR #452, opened 2025-04-07
feat: Improving wmsjobpolicy to contain pilots

Robin-Van-de-Merghel:
Implementation for #444 

Some interrogations about my understanding of this policy:

- We do need to check for a `job_id` associated with a pilot? (already implemented)
  - With `job_ids`, I assumed that a pilot can manage only one job at a time, do we?


Robin-Van-de-Merghel:
Needs tests

PR #454, opened 2025-04-07
fix (JobDB): #448 and extra checking for naive datetimes

### LINKED ISSUE #448: Is there a better way to handle enums in
@ryuwd As you've spend time with custom sqlalchemy types, could you take a look at if it's possible to avoid `_set_job_attributes_fix_value` in https://github.com/DIRACGrid/diracx/pull/447/commits/a08b7026c97bf08091f6c0ba972af263bc71a697.

Without it we end up inserting `0` and `1` into the DB in the unit test instead of `'True'` and `'False'`.

ryuwd:
TODO:
- [x] write docs on datetimes and timezones for diracx:
  - Everything is stored in UTC
  - Naive vs. aware and the differences between DIRAC to diracx
  - Avoiding catastrophes in datetime storage
- [x] simplify the API with sensible defaults assumed e.g. utc stored and retrieved
- [x] rebase

## resolution of #448 

The `UPDATE` in JobDB `set_job_attributes` has an extra layer to it involving a case statement per column which decides whether or not to update the column (i.e. if the caller said so). This is to support bulk updates where attributes for multiple jobs are updated in bulk, but the numbers of columns updated per job are not the same. 

It seems that the engine is not aware of how to correctly render the type inside the CASE, as it does not connect the dots to see which column and type is to receive the result of the CASE. When it comes to rendering the bind parameter, the type engine has no idea what to do, so it defaults to the python type passed to it (so if `True` or `False` is passed instead of `"True"` etc, it will just assume the user just wants to render a simple boolean inside a `CASE` statement.).

It seems the answer is to use `literal(value, type_=column_type)` to ensure the bind parameter for the new value is rendered with the type of the column to be updated.

## Knock on effects: Storage and retrieval of datetimes ensuring timezone awareness

Once you start being stricter on which column types you want to use to render your bind parameter values i.e. you render them as actual DateTime and not just a string/varchar etc: you cannot just pass strings anymore (e.g. this seems to be the case for sqlite). 

And, due to the timezone aware/non-aware stuff, some careful thought and changes to DateTime column types over all DB models will be needed to make sure DateTimes are treated in a consistent and sensible way.

What I've come up with is the `SmarterDateTime` TypeDecorator which accepts the following arguments

```python
    def __init__(
        self,
        # What timezone to use to store the datetime
        # None means stored in DB in local system timezone (beware, the DB is not queried for this...)
        # DRAGONS: the "local timezone" of the system connecting to the DB must match the timezone of the DB...
        # so to be sure that the timezones match there is the option to set it explicitly
        stored_tz: ZoneInfo | None = None, 
        returned_tz: ZoneInfo = ZoneInfo("UTC"), # what timezone is returned to the user
        stored_naive_sqlite=True, # Dialect specific switches if needed
        stored_naive_mysql=True,
        stored_naive_postgres=False,  # Forces timezone-awareness
    ):
```

Which enforces timezone-aware UTC datetimes to be returned out of the database, abstracting away the detail of converting between other timezones or tz-unaware/aware objects.

I've written it such that it doesn't matter if an INSERT or UPDATE passes a CEST, JST or whatever datetime, as in the end this type always converts to the form that the DB stores and returns UTC back out again.

I've only used this type so far with the JobDB as its not clear to me how the datetimes are stored in other DBs and Tables. For further discussion

cc @chrisburr, @chaen 

fixes #448


ryuwd:
The failing test is due to a date time column strictly requiring a python datetime type on render, which is a sensible ask. I think this leads to needing a `TypeDecorated` DateTime type that handles converting python strings to `datetime` objects when needed.

diracx-db/src/diracx/db/sql/utils/types.py
```diff
@@ -41,3 +39,37 @@ def process_result_value(self, value, dialect) -> bool:
...
+                raise ValueError(f"Unable to parse datetime string: {value}") from err
+
+        if not isinstance(value, datetime):
+            raise ValueError(f"Expected datetime or ISO8601 string, but got {value!r}")
+
+        return value
+
+    def process_result_value(self, value, dialect):
+        if isinstance(value, datetime) and not value.tzinfo:
+            value = value.replace(tzinfo=self._tz)
```
ryuwd, line 74: This is a potentially a landmine that needs to be checked

ryuwd: OK I've refactored the class a bit to do the time zone calculations with respect to whether the DB deals in local time or not

ryuwd:
Can't seem to reproduce the test failure on my mac or on lblhcbpr20...

diracx-db/src/diracx/db/sql/utils/types.py
```diff
@@ -41,3 +53,53 @@ def process_result_value(self, value, dialect) -> bool:
...
+        if self._stored_naive:
+            if isinstance(value, datetime) and not value.tzinfo:
+                # add back the timezone info
+                value = value.replace(tzinfo=self._tz)
+
+                if value.tzinfo is None:
+                    # if stored as a local time, add back the timezone info...
+                    value = value.replace(tzinfo=get_local_timezone())
+
+        return value
```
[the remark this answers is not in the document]
ryuwd: This has evolved a bit... see diff for details

ryuwd:
`tzlocal` seems to be missing: https://github.com/DIRACGrid/diracx/actions/runs/14368409595/job/40286869618?pr=454#step:8:3097

ryuwd:
added in https://github.com/DIRACGrid/container-images/pull/32

ryuwd:
Found that `datetime.astimezone()` does exactly what we need, so no need to install `tzlocal` anymore.

diracx-db/src/diracx/db/sql/utils/types.py
```diff
@@ -41,3 +40,98 @@ def process_result_value(self, value, dialect) -> bool:
...
+    naive form and back when needed.
+
+    """
+
+    impl = DateTime()
+    cache_ok = True
+
+    def __init__(
+        self,
+        stored_tz: ZoneInfo | None = None,
```
ryuwd, line 58:
```suggestion
        stored_tz: ZoneInfo | None = ZoneInfo("UTC"),
```

ryuwd:
Needs rebase

diracx-db/src/diracx/db/sql/utils/types.py
```diff
@@ -41,3 +40,98 @@ def process_result_value(self, value, dialect) -> bool:
...
+            )
+
+        # Check that we need to convert the timezone to match self._stored_tz timezone:
+        if self._stored_naive(dialect):
+            # if self._stored_tz is None, we use our local/system timezone.
+            stored_tz = self._stored_tz
+
+            # astimezone converts to the stored timezone (local timezone if None)
+            # replace strips the TZ info --> naive datetime object
+            value = value.astimezone(tz=stored_tz).replace(tzinfo=None)
```
AcquaDiGiorgio, line 106: Why do we need a `_stored_tz`? With just the `_returned_tz` it should be fine, no?

I might be wrong, but from what I see, the TZ of the date should become `_returned_tz` before inserting, because if `_stored_tz` and `_returned_tz` are somehow different, we might get illogical results.

AcquaDiGiorgio (commented):
I love this way of treating the time zone, it makes a lot of sense and is quite readable 👌

diracx-db/src/diracx/db/sql/utils/types.py
AcquaDiGiorgio: With further inspection, I see that it gets covered here:

```python
if self._stored_naive(dialect):
    # Here we add back the tzinfo to the naive timestamp
    # from the DB to make it aware again.
    if value.tzinfo is None:
        # we are definitely given a naive timestamp, so handle it.
        # add back the timezone info if stored_tz is set
        if self._stored_tz:
            value = value.replace(tzinfo=self._stored_tz)
        else:
            # if stored as a local time, add back the system timezone info...
            value = value.astimezone()
```

However, I still think it is redundant, we could just use the `_returned_tz` for both storing and returning.

AcquaDiGiorgio: I see, I see. I didn't know that was a case we needed to cover.

Then, I see nothing wrong in this PR 💯💯

ryuwd:
This should be rebased and re-checked - I will try to get around to it next week

fstagni (approved):

Issue #459, opened 2025-04-08
Logout on DiracX CLI not working


Robin-Van-de-Merghel:
When I login then logout:

```
(diracx-dev) fork@mac ~ % dirac logout
Error revoking the refresh token HttpResponseError("Operation returned an invalid status 'Bad Request'")
Removed credentials from [...]/credentials.json

Logout successful!
(diracx-dev) fork@mac ~ %
```

We can see a problem, a 400 bad request. 
The bug comes from the fact that the function `revoke_refresh_token` called [here](https://github.com/DIRACGrid/diracx/blob/2f22f5a3d6eafd79c068824d8fbecfe76eecd279/diracx-cli/src/diracx/cli/auth.py#L127) takes normally a `jti` as an input, not a refresh token.

I started a PR to change the router, so you can delete a token without having to find out what is your `jti`. But I saw then that log of tests depends on this endpoint, so I don't know what is the best:

- Reading the JWT from the CLI itself to preserve the endpoint
- Change the endpoint to have a more natural endpoint, using a refresh token

aldbr:
I am really surprised we didn't spot that issue earlier! Thanks 🙂 
It's probably best to change the endpoint to get the refresh token as input I would say.

Robin-Van-de-Merghel:
> It's probably best to change the endpoint to get the refresh token as input I would say.

That's what I think it's the better too, but we'll have to change ~8 tests that rely on this, I wanted to be sure before moving it.

PR #460, opened 2025-04-08
fix: Fixed logout router

Robin-Van-de-Merghel:
References #459 

Changed the router to remove `{jti}`, replaced with `refresh_token`


Robin-Van-de-Merghel:
Needs tests on cli. I'll do it tomorrow.

diracx-routers/src/diracx/routers/auth/management.py
```diff
@@ -74,7 +73,8 @@ async def revoke_refresh_token(
         subject = None
 
     try:
-        await revoke_refresh_token_bl(auth_db, subject, UUID(jti, version=4))
+        token_information = read_token(refresh_token, settings)
+        await revoke_refresh_token_bl(auth_db, subject, token_information["jti"])
```
aldbr, line 77:
```suggestion
        await revoke_refresh_token_bl(auth_db, settings, subject, refresh_tokens)
```

Could you use `read_token` from `diracx-logic`? To keep all the logic at the same place.

aldbr (commented):
Thanks for the PR!
Could you check whether we are correctly following and implementing https://datatracker.ietf.org/doc/html/rfc7009 please?

Robin-Van-de-Merghel:
> Thanks for the PR! Could you check whether we are correctly following and implementing https://datatracker.ietf.org/doc/html/rfc7009 please?

Just red the RFC, for the revocation response, we have:

>    The authorization server responds with HTTP status code 200 if the token has been revoked successfully or if the client submitted an invalid token.
> Note: invalid tokens do not cause an error response since the client cannot handle such an error in a reasonable way. Moreover, the purpose of the revocation request, invalidating the particular token, is already achieved.

Do we really follow it strictly? Most of our test relies on it, and we test for example if a user can revoke it's manager's token (which to me makes sens).  RFC tells us not to do that.

diracx-logic/src/diracx/logic/auth/token.py
```diff
@@ -337,6 +338,17 @@ def create_token(payload: TokenPayload, settings: AuthSettings) -> str:
     return encoded_jwt.decode("ascii")
 
 
+def read_token(payload: str, settings: AuthSettings) -> dict:
```
aldbr, line 341: I think there is already a similar function named `verify_dirac_refresh_token`, can't you just reuse it?

aldbr:
> Do we really follow it strictly? Most of our test relies on it, and we test for example if a user can revoke it's manager's token (which to me makes sens). RFC tells us not to do that.

I think it's best if we follow the RFCs as much as we can.
As you say, the `/revoke` endpoint would work for end users as they have their refresh token at hands, but would not be useful for "admins" that would need to revoke the refresh token of a given end user (he/she would not have access to the encoded refresh token of the user, but could get access to the `jti`).

May be the best solution would be to have both endpoints:
- `POST /revoke/` for end users to log out
- `DELETE /token/{jti}` for admins revoking a given end user's refresh token

What do you think?

Robin-Van-de-Merghel:
> What do you think?

I think it's a nice idea if we want to keep a clean logout for the user, and a good administrative endpoint. I must document it.

Robin-Van-de-Merghel:
Still need to take back old tests, and allow only admins.

Robin-Van-de-Merghel:
What do you think corresponds the best to admins?  In terms of security properties.

I think:

- `CS_ADMINISTRATOR`
- `SERVICE_ADMINISTRATOR`
- `SITE_MANAGER`
- `OPERATOR`

Maybe creating a `AdminAccessPolicy` would be nice by the way

diracx-routers/tests/auth/test_standard.py
aldbr, line 1: Why did you modify these tests?
I think we still want to test the `refresh-tokens/{jti}` endpoint.

diracx-routers/src/diracx/routers/auth/management.py
```diff
@@ -60,8 +57,32 @@ async def get_refresh_tokens(
...
+    subject: str | None = user_info.sub
+    if PROXY_MANAGEMENT in user_info.properties:
+        subject = None
+
+    try:
+        await revoke_refresh_token_by_refresh_token_bl(
+            auth_db, subject, refresh_token, settings
+        )
+    except ValueError:
+        print(f"{user_info.preferred_username} tried to revoke its token but failed.")
```
aldbr, line 79: You can probably instantiate a `logger` at the top of the module like in: https://github.com/DIRACGrid/diracx/blob/37525374c1d87c263de4f049bc86a98e095d0003/diracx-logic/src/diracx/logic/jobs/status.py#L53


And then use it instead of `print`: 
```suggestion
        logger.warning(f"{user_info.preferred_username} tried to revoke its token but failed.")
```

Even if at the moment we don't have a policy on how to use `logger` objects, `logging` is much more flexible than `print` (log granularity, output, thread-safe, filtering...)

diracx-logic/src/diracx/logic/auth/management.py
```diff
@@ -30,3 +32,16 @@ async def revoke_refresh_token(
 
     await auth_db.revoke_refresh_token(jti)
     return f"Refresh token {jti} revoked"
+
+
+async def revoke_refresh_token_by_refresh_token(
+    auth_db: AuthDB, subject: str | None, refresh_token: str, settings: AuthSettings
+) -> str:
+    """Revoke a refresh token. If a subject is provided, then the refresh token must be owned by that subject."""
+    token_information = read_token(
```
aldbr, line 41: Very minor comment but why not using `verify_dirac_refresh_token` now that you are using `read_tokens` within it?
I would say:
- either we just use `read_token` directly everywhere
- or we keep `read_token` "private" and we use it through `verify_dirac_access/refresh_token`

What do you think?

Robin-Van-de-Merghel: I think yes it's better to use `verify_dirac_refresh_token` as it is  its  purpose, and to make `read_token` general and use for specific cases

diracx-routers/tests/auth/test_standard.py
Robin-Van-de-Merghel: I added the old tests, and separated the new ones (with a little extra)

Robin-Van-de-Merghel:
> @Robin-Van-de-Merghel Can you rebase this?

Normally it's fine now

aldbr (dismissed):

diracx-routers/src/diracx/routers/auth/management.py
```diff
@@ -60,8 +61,34 @@ async def get_refresh_tokens(
...
+    user_info: Annotated[AuthorizedUserInfo, Depends(verify_dirac_access_token)],
+    settings: AuthSettings,
+    refresh_token: str,
+) -> str:
+    """Revoke a refresh token. If the user has the `proxy_management` property, then
+    the subject is not used to filter the refresh tokens.
+    """
+    subject: str | None = user_info.sub
+    if PROXY_MANAGEMENT in user_info.properties:
+        subject = None
```
aldbr, line 76: I wonder whether the access token + this bloc of code are required.

From what I can read in https://datatracker.ietf.org/doc/html/rfc7009#section-2.1, we are supposed to authenticate using the client credentials, and not using an access token, which might make sense as you might have a refresh token and an expired access token. 

> POST /revoke HTTP/1.1
> Host: server.example.com
> Content-Type: application/x-www-form-urlencoded
> Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW
>
> token=45ghiukldjahdnhzdauz&token_type_hint=refresh_token

- Could you try to add a unit test where you try to log out while having an expired access token?

Now if you come with your refresh token, then it generally means it belongs to you.
If you come with a stolen refresh token:
- if we also come with an access token as it's currently done: then we can potentially check whether they have the same `sub`, but is there any point?
- if we come with client creds: then we have no way to make sure that the refresh token belongs to the requester, but is that a problem?

What do you think?

aldbr (commented):
Could you also add the `revocation_endpoint` url to the `/.well-known/openid-configuration` route please? https://datatracker.ietf.org/doc/html/rfc8414#section-2

Thanks! :slightly_smiling_face:

diracx-routers/src/diracx/routers/auth/management.py
Robin-Van-de-Merghel: The part I don't really understand, is the one with "`we are supposed to authenticate using the client credentials`", as we do not have credentials? We use IAM, that manages credentials, to then authorize and create an access token.
Do you mean not checking if a token is expired or not, but still use access token?

For the rest:

- unit test with expired token: I'll do it
- verifying if a token is really yours: in a way we do not care, because revoking a token as a stealer destroys our possibilities to later steal

diracx-routers/tests/auth/test_standard.py
```diff
@@ -852,6 +852,65 @@ async def test_revoke_refresh_tokens_token_manager(
...
+
+    # Get all refresh tokens, expect no refresh token
+    r = test_client.get(
+        "/api/auth/refresh-tokens",
+        headers={"Authorization": f"Bearer {normal_user_access_token}"},
+    )
+
+    assert r.json() == []
+
+    # Normal user tries to delete his/her RT: should work
```
aldbr, line 901: Very minor and last comment (before having a working client generation :sweat_smile: )

```suggestion
    # Normal user tries to delete a valid RT using the wrong client id
```

aldbr (approved):

Issue #417, opened 2025-03-26
Open access and require auth not working at all


Robin-Van-de-Merghel:
# Explanation

I have a question about rights:

`open_access` documentation says: 

> The presence of a token will still be checked if the router has `require_auth` to `True`.

That implies that having `open_access` and `require_auth` set to `False` will not verify a JWT, *so it won't check a `Authorization Header`* (it is what I want with Pilots registrations, because they don't have jwt at the first time). 

It is confirmed in the service documentation:

---

> Some routes do not need access permissions, like the authorization ones, in which case they can be marked as such:

```python
from .access_policies import open_access

@open_access
@router.get("/")
async def serve_config():
    ...
```

---

When I do everything, if I add a Policy, the policy works very well. If I remove the policy, it requires a auth header, that's normal.  But when I add `open_access` and/or `require_auth=False`, I still need auth header, and  to be precise, I still need a valid JWT.

Is it a bug or a feature? (Or just me)

# Try to fix

I have crawled into the code, and have seen that `open_access` sets `f.diracx_open_access = True`. And `diracx_open_access` is not used at all except in tests:

![Image](https://github.com/user-attachments/assets/87994028-d372-4a21-8efb-351aef167940)

The bug seems to come from `factory.py`:

```py
        # Add the router to the application
        dependencies = []
        if isinstance(router, DiracxRouter) and router.diracx_require_auth: # CREATES BUG
            dependencies.append(Depends(verify_dirac_access_token))         # CREATES BUG  
        # Most routers are mounted under /api/<system_name>
        path_root = getattr(router, "diracx_path_root", "/api")
        app.include_router(
            router,
            prefix=f"{path_root}/{system_name}",
            tags=[system_name],
            dependencies=dependencies,
        )
```

By commenting this part, we won't need a authorization header anymore. So I guess the problem comes from the `and router.diracx_require_auth:` condition.

# To try it

My code:

```py
# [Imports]

router = DiracxRouter(require_auth=False) # Does not create an open access

@open_access # Does not create an open access
@router.post("/register")
async def search(
    # pilot_db: PilotAgentsDB,  # Not Useful for the issue
    # check_permission: RegisteredPilotAccessPolicyCallable,
    pilot_reference: str = "ref",
):
    
    # await check_permission(pilot_db=pilot_db) # Works pretty well
    
    return pilot_reference
````

###  Without Header

Here I run a `POST` request to my endpoint, without any authorization header:

```json
{
[...]

PR #468, opened 2025-04-09
feat: Improving wmsjobpolicy to contain pilots

Robin-Van-de-Merghel:
Reopened*

Implementation for #444 

Some interrogations about my understanding of this policy:

- We do need to check for a job_id associated with a pilot? (already implemented)
- With job_ids, I assumed that a pilot can manage only one job at a time, do we?


chrisburr:
> With job_ids, I assumed that a pilot can manage only one job at a time, do we?

A pilot can potentially be running many jobs at the same time (via `PoolCE` in legacy DIRAC).

diracx-routers/src/diracx/routers/jobs/access_policies.py
```diff
@@ -41,16 +46,50 @@ async def policy(
...
+
+            assert (
+                pilot_db
+            ), "pilot_db is a mandatory parameter when using a pilot action"
+            assert job_ids, "job_ids has to be defined"
+            pilot_info = user_info  # For semantic
+
+            # Syntax to avoid code duplication
+            if {GENERIC_PILOT, LIMITED_DELEGATION} & set(pilot_info.properties):
+                # Get his informations
```
fstagni, line 66:
```suggestion
                # Get its information
```
?

Robin-Van-de-Merghel:
Facing the problem from #417, set `require_auth` to `False`.

Robin-Van-de-Merghel:
Router won't pass the tests: we need to have in the factory (or somewhere else) a pilot token generator (but not a "`pilot_client`" as `normal_user_client`, because we can't stack multiple contexts).

PR #469, opened 2025-04-09
fix: job admins cannot see jobs from other users

aldbr:
Currently users with `JOB_ADMINISTRATORS` properties cannot see jobs of other users because of: https://github.com/aldbr/diracx/blob/a37ff0db0dab93ce90d44938310ba35ed1f1cbd9/diracx-logic/src/diracx/logic/jobs/query.py#L49-L60

I don't know if this is the best way of handling that problem, because I see some `TODO` notes like `# TODO: Apply all the job policy stuff properly using user_info`.

I am coming with some unit tests soon.

### DIFF
--- a/diracx-logic/src/diracx/logic/jobs/query.py
+++ b/diracx-logic/src/diracx/logic/jobs/query.py
@@ -24,7 +24,7 @@ async def search(
     job_db: JobDB,
     job_parameters_db: JobParametersDB,
     job_logging_db: JobLoggingDB,
-    preferred_username: str,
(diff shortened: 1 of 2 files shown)


aldbr:
Update: the test was trickier than I would have expected, I checked a few weeks ago but I would need more time to make this work, as the `ClientFactory` cannot currently change its configuration.
As the PR might block the other ones and that I won't work on it in the next weeks/month, I opened an issue to remind me adding a test for it later.

fstagni (approved):

PR #472, opened 2025-04-10
Lock cache

chaen:


diracx-routers/src/diracx/routers/otel.py
```diff
@@ -46,6 +48,29 @@ class OTELSettings(ServiceSettingsBase):
...
+            try:
+                enable_otel = bool(int(c) % 2)
+                break
+            except ValueError:
+                pass
+    except Exception as exe:
+        logger.exception(exe)
+
+    logger.info(f"Randomly enabling OTEL {enable_otel=}")
+    return enable_otel
```
chrisburr, line 71:
```suggestion
    enable_otel = bool(int(next(filter(str.isdigit, socket.gethostname() + "0"))) % 2)
    logger.info(f"Randomly enabling OTEL {enable_otel=}")
    return enable_otel
```

chaen: a hell lot less readable :laughing:

chaen: and it doesn't go backward

chrisburr:
I think this one can be closed right?

Issue #471, opened 2025-04-10
Implement proper health check endpoint


chaen:
Pods are marked as ready when they can reach the `/api/docs` endpoint.
This is not good because it does not use the `ConfigSource`, so a pod is marked as ready before we can make sure it can load the config. 
We should have proper health endpoint

https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check

PR #473, opened 2025-04-10
Add health checks

chrisburr:


diracx-routers/src/diracx/routers/health/probes.py
```diff
@@ -0,0 +1,51 @@
...
+    Checks if at least the configuration is loaded and the AuthDB database
+    connection is available.
+    """
+    if not any(vo_registry.Users for vo_registry in config.Registry.values()):
+        raise HTTPException(status_code=503, detail="No users in registry")
+    try:
+        await auth_db.ping()
+    except Exception as e:
+        raise HTTPException(status_code=503, detail="AuthDB ping failed") from e
+    return JSONResponse(content={"status": "startup complete"})
```
aldbr, line 51: What's the difference between `GET /startup` and `GET /ready`? They look similar, is that expected?

aldbr (commented):
You are not following https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check because it is `expired and archived`, aren't you?

Robin-Van-de-Merghel:
(to link it) #471

diracx-routers/src/diracx/routers/health/probes.py
aldbr: Ah yes, one if for the initialization, the other one is continuously check, but the same details need to be checked. As it is very short code, it's fine to be duplicated.

chrisburr: Yeah abstracting it felt uglier than the duplication. Maybe over time they diverge to startup being heavier than ready.

PR #478, opened 2025-04-13
fix: shun naive datetimes in CODING_CONVENTIONS.md

ryuwd:


chrisburr:
There is at least one more occurance further down the file.

ryuwd:
> There is at least one more occurance further down the file.

That is in the `Do not` column i.e. do not use without `tz=...` - did you mean somewhere else?

fstagni (approved):

Issue #461, opened 2025-04-09
Remove RefreshTokens.preferred_username


chrisburr:
We're fairly sure this isn't needed and could be added from the CS instead.

PR #485, opened 2025-04-16
fix: remove preferred username from the RefreshToken table

aldbr:
Closes https://github.com/DIRACGrid/diracx/issues/461


diracx-logic/src/diracx/logic/auth/token.py
```diff
@@ -212,7 +212,6 @@ async def get_oidc_token_info_from_refresh_flow(
         # The sub attribute coming from the DB contains the VO name
         # We need to remove it as if it were coming from an ID token from an external IdP
         "sub": sub.split(":", 1)[1],
-        "preferred_username": refresh_token_attributes["PreferredUsername"],
```
chrisburr, line 215: Isn't this still needed but should be taken from the config rather than the refresh token?

aldbr: Yes indeed, and the following piece of code seems to do the job already:

https://github.com/DIRACGrid/diracx/blob/c5a8b605bdda1fff6ca76ca6c1d7da860d60fd08/diracx-logic/src/diracx/logic/auth/token.py#L293-L294

chrisburr: I haven't looked closely but isn't that just for getting the initial token after doing a flow against IAM?

For later getting a new access token using the refresh token I don't see how the `preferred_username` is populated.

chrisburr: Indeed I was just mixed up!

chrisburr (approved):

PR #486, opened 2025-04-17
Use uuid7

chrisburr:
This pull request introduces a migration from `uuid4` to `uuid7` for generating unique identifiers. The change improves the temporal ordering of UUIDs, which can be beneficial for database indexing. I won't go into why and suggest anybody unaware of UUID v7 reads https://uuid7.com/.

At the moment it isn't available in the Python standard library (having only recently been accepted and still having "draft" status) so the `uuid-utils` library is used. There is a decent chance it'll be included in Python 3.14 later this year.

To see that this isn't a theoretical improvement, you can see pretty clearly when I switched to this in LHCb's DB monitoring (see below). This doesn't give a complete story as the DB is still poisened by the existing uuid4 records which is why there is a tail effect.

Screenshot 2025-04-17 at 08 30 05

It also halved the average duration of `exchangeProxyForToken` calls and significantly improved the tail latency:

With uuid4:

Screenshot 2025-04-17 at 08 49 39

With uuid7 (mostly, still only 8 hours after switching):

Screenshot 2025-04-17 at 08 55 33

### DIFF
--- a/diracx-db/pyproject.toml
+++ b/diracx-db/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
     "opensearch-py[async]",
     "pydantic >=2.10",
     "sqlalchemy[aiomysql,aiosqlite] >= 2",
+    "uuid-utils",
 ]
 dynamic = ["version"]
 
--- a/diracx-db/src/diracx/db/sql/auth/db.py
+++ b/diracx-db/src/diracx/db/sql/auth/db.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 import secrets
-from uuid import UUID, uuid4
 
 from sqlalchemy import insert, select, update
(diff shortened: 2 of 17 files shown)


aldbr (commented):
Awesome!
Would it make sense to have a utility function somewhere, named `get_uuid()` for instance, to avoid having to change `uuid7()` everywhere as we will probably have to do it again once `uuid7` will be part of the standard library?

chrisburr:
> Would it make sense to have a utility function somewhere, named get_uuid() for instance, to avoid having to change uuid7() everywhere as we will probably have to do it again once uuid7 will be part of the standard library?

I'm inclined to not do as we'll also have to change the type of everythng as well

aldbr:
Indeed :sweat_smile:

aldbr (approved):

PR #494, opened 2025-04-24
Fixed current exceptions in Dirac Core

Robin-Van-de-Merghel:
Fixed the current errors (bad use of f-strings, wrong refactoring, migrating to `RuntimeError`).

# Suggestion

In this PR, I can also add something to avoid duplication:

```py
class GenericError(DiracError):
    pattern = "Error: %s"

    def __init__(self, data: dict[str, str], detail: str | None = None):
        self.data = data
        self.detail = detail

        parts = [f"({key}: {value})" for key, value in data.items()]
        message = pattern % {' '.join(parts)}
        if detail:
            pattern += f": {detail}"

        super().__init__(message)
```

The idea: you extend this class with your proper pattern, to avoid duplicating. It also helps having a flexibility on the parameters we give to the error. Example:

```py
class PilotNotFoundError(GenericError):
    pattern = "Pilot %s not found"

raise MyPersonalizedError(data={
    "pilot_stamp": "a-stamp"
})
# Raises "Pilot (stamp: a-stamp) not found"


raise MyPersonalizedError(data={
    "pilot_ref": "a-ref"
}, detail="I don't know why")
# Raises "Pilot (pilot_ref: a-ref) not found: I don't know why"
```

The last example shows that you can re-use it with another detail (stamps, references, vos, ...). It also ensures that every error is quite pre-defined, and not diagonally different between each others.

__I don't know if it would be nice considering the rest of the errors?__

### DIFF
--- a/diracx-core/src/diracx/core/exceptions.py
+++ b/diracx-core/src/diracx/core/exceptions.py
@@ -52,54 +52,55 @@ class InvalidQueryError(DiracError):
     """It was not possible to build a valid database query from the given input."""
 
 
-class TokenNotFoundError(Exception):
+class TokenNotFoundError(DiracError):
     def __init__(self, jti: str, detail: str | None = None):
         self.jti: str = jti
-        super().__init__(f"Token {jti} not found" + (" ({detail})" if detail else ""))
+        super().__init__(f"Token {jti} not found" + (f" ({detail})" if detail else ""))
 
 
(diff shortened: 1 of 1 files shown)


fstagni:
I think it's a nice idea, but out of simplicity I would keep it as it is.

fstagni (approved):

PR #496, opened 2025-04-25
[HotFix]: Bad error handling can lead to a 500 error

Robin-Van-de-Merghel:
Changed the error handling in the fixed logout router: [in these lines](https://github.com/Robin-Van-de-Merghel/diracx/blob/4c72d9a5225d6e8fb930d14fafe808c3261f94a4/diracx-logic/src/diracx/logic/auth/utils.py#L193-L205).

This led if the token is badly shaped to a crash in `verify_dirac_access_token`, because we catch only `JoseError`, but we raise `ValueError` which is not caught.

I found this bug doing [CI in another PR](https://github.com/DIRACGrid/diracx/actions/runs/14658762686/job/41138418074?pr=421), but we did not catch this elsewhere because this is not tested everywhere.

### DIFF
--- a/diracx-logic/src/diracx/logic/auth/utils.py
+++ b/diracx-logic/src/diracx/logic/auth/utils.py
@@ -8,7 +8,6 @@ import secrets
 import httpx
 from authlib.integrations.starlette_client import OAuthError
 from authlib.jose import JsonWebKey, JsonWebToken
-from authlib.jose.errors import DecodeError
 from authlib.oidc.core import IDToken
 from cachetools import TTLCache
 from cryptography.fernet import Fernet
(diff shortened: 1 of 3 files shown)


aldbr:
Why did we miss it in the unit tests?

Robin-Van-de-Merghel:
> Why did we miss it in the unit tests?

Unit tests does not test the case where we want to access a resource where you need a token, and that your "token" is a random Bearer. 

If you have `Authorisation: sdsdsdsdsd` it will throw a 400, but we don't test the case where `Authorisation: Bearer randomStringButNotToken` (on the unit tests)

aldbr (approved):

PR #497, opened 2025-04-28
Adding a context for bad DB state error handling

Robin-Van-de-Merghel:
This class could help identify easily a bad DB state, and help reporting it.

The idea:

```py
with DBStateAssertation(self.conn, [PilotNotFoundError, SecretNotFoundError]):
  # This body should NOT raise PilotNotFoundError nor SecretNotFoundError
  # If it does, it logs it, and raise a different error (DBInBadStateError) to avoid returning a 404 error
  # -> It will return a 500 error, which is wanted because it is an internal error
  
  # It will also rollback changes
  await self.a_db_function_that_should_not_raise_these_errors()
```

I opened this PR because I use this mechanic a lot to make sure that everything is fine. See examples:

- https://github.com/DIRACGrid/diracx/blob/c7b196dcf7791c5808876d510d3b8eea432e802b/diracx-db/src/diracx/db/sql/pilot_agents/db.py#L73-L80
- https://github.com/DIRACGrid/diracx/blob/c7b196dcf7791c5808876d510d3b8eea432e802b/diracx-db/src/diracx/db/sql/pilot_agents/db.py#L114-L123
- https://github.com/DIRACGrid/diracx/blob/c7b196dcf7791c5808876d510d3b8eea432e802b/diracx-db/src/diracx/db/sql/pilot_agents/db.py#L145-L153
- etc.

### DIFF
--- a/diracx-db/src/diracx/db/sql/utils/functions.py
+++ b/diracx-db/src/diracx/db/sql/utils/functions.py
@@ -1,12 +1,18 @@
 from __future__ import annotations
 
 import hashlib
+import logging
 from datetime import datetime, timedelta, timezone
 from typing import TYPE_CHECKING
 
 from sqlalchemy import DateTime, func
+from sqlalchemy.ext.asyncio import AsyncConnection
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.sql import expression
+
+from diracx.db.exceptions import DBInBadStateError
+
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from sqlalchemy.types import TypeEngine
@@ -140,3 +146,54 @@ def substract_date(**kwargs: float) -> datetime:
 
 def hash(code: str):
     return hashlib.sha256(code.encode()).hexdigest()
+
+
+class DBStateAssertion:
+    """Context manager to ensure the integrity of a database transaction,
+    specifically by rolling back any changes if certain expected exceptions
+    occur—and raising a standardized error to indicate the database is in a bad state.
+
+    Example:
+    with DBStateAssertation(self.conn, [PilotNotFoundError, SecretNotFoundError]):
+        # Some code that should not raise any error filled previously
+        # Else, raises an error
+
+    Note: Will rollback changes.
+
+    """
+
+    def __init__(
+        self, conn: AsyncConnection, exceptions: list[type[Exception]]
+    ) -> None:
+        self.exceptions = exceptions
+        self.conn = conn
+
+    async def __aenter__(self):
+        pass
+
+    async def __aexit__(self, exc_type, exc_value, exc_tb):
+        if exc_type is None:
+            # No exception occurred
+            return False
+
+        # If we get here, an error occurred, so we rollback changes
+        # (Rollback will be done by the fastapi application itself)
+        # (See routes and db documentation)
+        message = (
+            f"Rolling back DB transaction due to: {exc_type.__name__}: {exc_value}"
+        )
+
+        # Check if the exception is among the expected ones
+        if any(issubclass(exc_type, exc) for exc in self.exceptions):
+            logger.warning(f"{message}. This error was expected.")
+            raise DBInBadStateError(
+                "This error may NOT have been raised. "
+                "Please report this at https://github.com/DIRACGrid/diracx/issues"
+            ) from exc_value
+
+        # Unexpected exception; raise an unexpected error
+        logger.warning(f"{message}. This error was NOT expected.")
+        raise DBInBadStateError(
+            f"Unexpected error ({exc_type.__name__}): {exc_value}. "
+            "Please report this at https://github.com/DIRACGrid/diracx/issues"
+        ) from exc_value
--- a/diracx-db/src/diracx/db/exceptions.py
+++ b/diracx-db/src/diracx/db/exceptions.py
@@ -3,3 +3,7 @@ from __future__ import annotations
 
 class DBUnavailableError(Exception):
     pass
+
+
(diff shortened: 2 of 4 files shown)


diracx-db/src/diracx/db/sql/utils/functions.py
```diff
@@ -109,3 +115,49 @@ def substract_date(**kwargs: float) -> datetime:
...
+            raise DBInBadStateError(
+                "This error may NOT have been raised. "
+                "Please report this error: https://github.com/DIRACGrid/diracx/issues"
+            ) from exc_value
+
+        # Not an expected exception; raise an unexpected error
+        raise DBInBadStateError(
+            "This should NOT happen. A miscellaneous error occured. "
+            "Please report this error: https://github.com/DIRACGrid/diracx/issues"
+        ) from exc_value
```
fstagni, line 163:
```suggestion
      raise DBInBadStateError(
          f"Unexpected error ({exc_type.__name__}): {exc_value}. "
          "Please report this at https://github.com/DIRACGrid/diracx/issues"
      ) from exc_value
```

diracx-db/src/diracx/db/sql/utils/functions.py
```diff
@@ -109,3 +115,49 @@ def substract_date(**kwargs: float) -> datetime:
...
+            # No exception occurred
+            return False
+
+        # If we get here, an error occured, so we rollback changes
+        await self.conn.rollback()
+
+        # Check if the exception is among the expected ones
+        if any(issubclass(exc_type, exc) for exc in self.exceptions):
+            logger.error(
+                f"An error occured with the db, exception message: {exc_value}"
```
fstagni, line 152:
```suggestion
                f"An error occurred with the db, exception message: {exc_value}"
```

diracx-db/src/diracx/db/sql/utils/functions.py
```diff
@@ -109,3 +115,49 @@ def substract_date(**kwargs: float) -> datetime:
 
 def hash(code: str):
     return hashlib.sha256(code.encode()).hexdigest()
+
+
+class DBStateAssertation:
+    """Class to handler context where we should not raise any excepted error.
```
fstagni, line 121:
```suggestion
    """Context manager to ensure the integrity of a database transaction,
    specifically by rolling back any changes if certain expected exceptions occur—and raising a standardized error to indicate the database is in a bad state.
```

diracx-db/src/diracx/db/sql/utils/functions.py
```diff
@@ -109,3 +115,49 @@ def substract_date(**kwargs: float) -> datetime:
 
 def hash(code: str):
     return hashlib.sha256(code.encode()).hexdigest()
+
+
+class DBStateAssertation:
```
fstagni, line 120:
```suggestion
class DBStateAssertion:
```
?

diracx-db/src/diracx/db/sql/utils/functions.py
```diff
@@ -109,3 +115,49 @@ def substract_date(**kwargs: float) -> datetime:
...
+    async def __aenter__(self):
+        pass
+
+    async def __aexit__(self, exc_type, exc_value, exc_tb):
+        if exc_type is None:
+            # No exception occurred
+            return False
+
+        # If we get here, an error occured, so we rollback changes
+        await self.conn.rollback()
```
fstagni, line 147:
```suggestion
        await self.conn.rollback()
        logger.warning(f"Rolling back DB transaction due to: {exc_type.__name__}: {exc_value}")
```

diracx-db/src/diracx/db/sql/utils/functions.py
```diff
@@ -140,3 +146,53 @@ def substract_date(**kwargs: float) -> datetime:
...
+    async def __aexit__(self, exc_type, exc_value, exc_tb):
+        if exc_type is None:
+            # No exception occurred
+            return False
+
+        # If we get here, an error occurred, so we rollback changes
+        message = (
+            f"Rolling back DB transaction due to: {exc_type.__name__}: {exc_value}"
+        )
+        await self.conn.rollback()
```
aldbr, line 183: I guess you don't need to rollback manually here, as it's supposed to be done in https://github.com/DIRACGrid/diracx/blob/a1e542c97ab1b297259e4e471249fee4c04a77d9/diracx-db/src/diracx/db/sql/utils/base.py#L208-L217
 
```suggestion
```

chaen:
> I don't know if it's useful for everybody. I know I would use it, but if I'm the only one maybe it's not relevant

I confess having a hard time understanding what is the benefit of that ? All the transaction model is already handled under the hood, so I don't understand what is this thing doing for me ?

Robin-Van-de-Merghel:
> I confess having a hard time understanding what is the benefit of that ? All the transaction model is already handled under the hood, so I don't understand what is this thing doing for me ?

If I understood well, what the transaction model catches is if there were any error raised, and if so, it rollbacks.

My goal with this is to add assertions throughout the db : each time I insert or update, I verify if there's the good amount of data changed. For example:

1. I use a pilot secret.
2. Secret used too much time, so I delete it. 
3. While deleting it, 0 secret has been deleted, this is very bad and **should not** happen, and I don't want the router to catch it as a 404 error or even return a 204.

Normally the update always works and always update the right amount of secrets, but if someday because of a bug we don't delete any secret, we won't catch it.

---

So what's it's adding, it's just a wrapper to assert some states, because [adding assertions could make the code cleaner](https://swehb.nasa.gov/display/SWEHBVC/9.15+Safe+Transitions)

PR #499, opened 2025-04-30
feat: introduce jwks

aldbr:
Goes from using a single key to sign/verify tokens to setting up a jwks following [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517).

Introduces 2 new `diracx-logic` commands:
- generate a new key and rotate the JWKS: the new key can sign/verify tokens, the old one(s) can only verify tokens.
- delete a given key from the JWKS

Note: I have switched from `authlib` to `joserfc` to manipulate `JWK` and `JWT` as recommended in the official documentation: https://docs.authlib.org/en/latest/jose/index.html#jose-guide

It comes with the following PRs:
- https://github.com/DIRACGrid/DIRAC/pull/8188
- https://github.com/DIRACGrid/container-images/pull/33
- https://github.com/DIRACGrid/diracx-charts/pull/151

**Migration Plan** (LHCb-only I think)

1. Get the private key
```bash
kubectl exec -it diracx-<pod_id> -- cat /signing-key/rsa256.key > rsa256.key
```

2. Import in a JWKS 
```python
import json
from joserfc.jwk import RSAKey, KeySet
from uuid_utils import uuid7


with open("rsa256.key") as f:
    key = RSAKey.import_key(
        f.read(),
        {
            "key_ops": ["sign", "verify"],
            "alg": "RS256",
            "kid": uuid7().hex, 
        }
    )
    
jwks = KeySet([key])

with open("jwks.json", "w") as f:
    json.dump(jwks.as_dict(private=True), f, indent=2)
```

3. Rotate the key (generate a new one)

```bash
python -m diracx.logic rotate-jwk --jwks-path jwks.json
```

4. Modify `values.yaml` to disable to keystore generation:

```bash
helm get values diracx -o yaml > values.yaml
```

```yaml
initKeyStore:
  enabled: false
```

5. Generate the secret

```bash
kubectl create secret generic diracx-jwks \
  --namespace=$namespace \
  --from-file=jwks.json \
```

Once the update is done, the content of the secret is expected to be copied in `jwks.json` and read by `diracx`.


aldbr:
Of course we have a chicken-and-egg issue because the repos depend on each other.
- I will make that PR compatible with the current approach (single key), this is only required for the tests
- I will add `authlib` back to `container-images`, along with `joserfc` for now

From there, we can:
- Merge https://github.com/DIRACGrid/container-images/pull/33
- I make a quick test with the certification environment by applying this branch, just to make sure the migration plan works fine.
- I follow the same procedure to create the keystore secret in lhcb-diracx instances to be ready.
- Merge this PR
- Merge https://github.com/DIRACGrid/diracx-charts/pull/151 and https://github.com/DIRACGrid/DIRAC/pull/8188 (they should pass the CI)
- I will then remove `authlib` from the dependencies in `container-images` and the compatibility with the single key in `diracx`.

The deployed instances should fetch the keystore instead of the single key and that should work transparently.

PR #503, opened 2025-05-05
RepoReview:  spellchecker

martynia:
Spellchecker, excluding generated and test code.


diracx-routers/src/diracx/routers/factory.py
```diff
@@ -19,7 +19,7 @@
 import dotenv
 from cachetools import TTLCache
 from fastapi import APIRouter, Depends, FastAPI, HTTPException, Request, status
-from fastapi.dependencies.models import Dependant
+from fastapi.dependencies.models import Dependent
```
fstagni, line 22: Nope: https://github.com/fastapi/fastapi/blob/master/fastapi/dependencies/models.py#L15

fstagni (changes requested):

PR #509, opened 2025-05-05
feat: added definitions of summary tables (JobDB and PilotAgentsDB)

fstagni:
Following https://github.com/DIRACGrid/DIRAC/pull/8181


fstagni:
Replaced by #555

PR #514, opened 2025-05-06
feat: add [redundant-expr, truthy-bool] to mypy

martynia:


martynia:
```python
def _apply_default_scheme(value: str) -> str:
 """Applies the default git+file:// scheme if not present."""
 if isinstance(value, str) and "://" not in value:
     value = f"git+file://{value}"
 return value


class AnyUrlWithoutHost(AnyUrl):
    _constraints = UrlConstraints(host_required=False)


ConfigSourceUrl = Annotated[AnyUrlWithoutHost, BeforeValidator(_apply_default_scheme)]
```

@aldbr MyPy complains about  code listed above because `value` is declared as `str` which makes `isinstance` test redundant. Pydantic manual hints that types in this case should be generic (they have Any). This doesn't work for us, because AnyUrl scheme attribute is read-only:
```python
elif isinstance(value, AnyUrl):
     value.scheme = "git+file" 
```

so this doesn't work and w/o it the validator would let in AnyUrl  with any scheme which denies its purpose.

PR #518, opened 2025-05-07
Use pixi

chrisburr:


fstagni:
> More information will be included in the upcoming PR for the developer docs.

Can you add that documentation together with this PR?

chaen:
> > More information will be included in the upcoming PR for the developer docs.
> 
> Can you add that documentation together with this PR?

we have a massive doc PR coming up, it's probably easier to keep it there, but promise, it will be in it :-D

.github/workflows/extensions.yml
aldbr, line 1: I personally don't mind getting rid of `extension.yaml`, but wasn't it supposed to be used as an "example" for the communities building their extensions?

chrisburr: My suggestion is that we duplicate some of the CI config and have a `extensions/gubbins/.github/workflows/main.yml` file containing the config for a plain extension rather than having a mixed state where the CI config has special cases due to being part of the main diracx repo.

In the medium term would could make the deployment CI push it to a dedicated example repo where the CI would actually run without any diracx special-ness.

The same applies to a `extensions/gubbins/pixi.toml` file.

chaen (approved):

Issue #349, opened 2024-12-18
Publish `gubbins:dev`


aldbr:
So that we can reuse in `diracx-web`

PR #527, opened 2025-05-09
feat: deploy gubbins images

aldbr:
Closes https://github.com/DIRACGrid/diracx/issues/349

- This would avoid duplicating the creation of the images in https://github.com/DIRACGrid/diracx-web/blob/main/.github/workflows/gubbins-test.yml.
- And this would also help front end developers when they have to debug the cypress tests not passing the CI. (e.g. https://github.com/DIRACGrid/diracx-web/pull/335)

I could potentially create a reusable workflow to avoid duplicating some code between `deployment.yaml` and `extension.yaml` but this would also add some complexity to `extension.yaml`. As it can act as an example of CI, I feel like it should stay clear and simple and that we can afford that small duplication (I've tried to minimize it). Please let me know if you have a different opinion.

Here I export both `gubbins/client` and `services`, but I think we would just need to use `gubbins/services`.

Unless you disagree, or if you have a better plan, I will make a quick test on my repo first, and if everything goes well I undraft the PR.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -219,3 +219,109 @@ jobs:
           platforms: linux/amd64,linux/arm64
           build-args: |
             EXTRA_PACKAGES_TO_INSTALL=git+https://github.com/DIRACGrid/DIRAC.git@integration
+
+# -------------------------------------------------------------------------------------------------
+# Gubbins deployment
+# -------------------------------------------------------------------------------------------------
+
+  build-gubbins-wheels:
+    name: Build wheels
+    runs-on: "ubuntu-latest"
+    needs: deploy-pypi
+    if: ${{ needs.deploy-pypi.outputs.create-release == 'true' }}
+    defaults:
+      run:
+        # We need extglob for REFERENCE_BRANCH substitution
+        shell: bash -l -O extglob {0}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: Installing dependencies
+        run: |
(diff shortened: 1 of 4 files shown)


aldbr:
So I applied https://github.com/DIRACGrid/diracx/commit/b830754f89565ef6ab0a24feebff8dfa35b85f73 to test the content of the PR on my fork.

It seems to work:
- `extensions.yaml`: https://github.com/aldbr/diracx/actions/runs/14975769023 (`gubbins-cli` unit tests are failing but it's independent from this PR, it's explained in https://github.com/DIRACGrid/diracx/issues/532)
- `deployment.yaml` (PR): https://github.com/aldbr/diracx/actions/runs/14975244414
- `deployment.yaml` (workflow dispatch): https://github.com/aldbr/diracx/actions/runs/14975547633


Note: `extensions.yaml` cannot pass this CI because it requires an existing `gubbins/services:dev` image, which does not exist at the moment.

aldbr:
I've applied the latest changes, and made things a bit differently in `main.yaml`: instead of relying on a potential `dev` image that would be stored in the registry (`diracx-services:dev` exists but not `gubbins-services:dev`), I build `dev` from scratch in the CI for both extensions.

It allows testing potential changes that could occur in the `Dockerfile`s (`diracx` and `gubbins`)

aldbr:
Looks like I don't have enough space to run `minio` as I am now loading the diracx docker image both locally and in `kind`:
https://github.com/DIRACGrid/diracx/actions/runs/15189075216/job/42716849956#step:13:1171

I am trying to add a `prune-loaded-images` option to `run_demo.sh` to remove the local image once loaded in `kind`, let's see how it goes :crossed_fingers:

PR #534, opened 2025-05-13
Sandbox store cleaning

chrisburr:
Needs #518


diracx-logic/tests/logic/test_sandboxes.py
```diff
@@ -0,0 +1,154 @@
...
+async def sandbox_settings(
+    test_sandbox_settings,
+) -> AsyncGenerator[SandboxStoreSettings, None]:
+    """Create a fake sandbox settings."""
+    async with test_sandbox_settings.lifetime_function():
+        yield test_sandbox_settings
+
+
+@pytest.fixture()
+def frozen_time() -> Generator[freezegun.FreezeGun, None]:
```
ryuwd, line 57: I wonder if this fixture kicks in before the databases are created and e.g. other fixtures such as the one which insert valid jobs into the test DBs?

chrisburr: There is no guarantees around the order by default:

https://docs.pytest.org/en/stable/reference/fixtures.html#fixture-instantiation-order

The solution would be to make the `valid_job_ids` depend on the frozen time or use a function instead of a fixture and just call it inside the test function.

PR #511, opened 2025-05-05
Pilot logging

martynia:

PR #550, opened 2025-05-22
Robin pilot logging

Robin-Van-de-Merghel:
Updates to #511 , rebased and linked to #421


Robin-Van-de-Merghel:
To review this code, this is a summary of what's new compared to the pilot authentification:

- `diracx-routers/src/diracx/routers/pilots/logging.py`: Upload messages
- `diracx-routers/src/diracx/routers/pilot_management/query.py`: There's `query_logs` which helps querying logs with the same idea as job queries
- `diracx-logic/src/diracx/logic/pilots/logging.py`: Tools to update and query logs
- Modified `diracx-testing/src/diracx/testing/mock_osdb.py` and `util.py` to test OSDB with new types: `date_nanos`. Also added support for queries: returned total (for `Content-Range`).
- `diracx-core/src/diracx/core/models.py`: Added logs
- `diracx-db/src/diracx/db/*.py`: Added logs and updated debug (when we add logs)

martynia:
I'm wondering why you removed the original line number from the `send_message` when storing a document. You added a timestamp which is useful, but it might be problematic when sorting messages. The messages should be presented in the order they were emitted by a client.

Robin-Van-de-Merghel:
> I'm wondering why you removed the original line number from the `send_message` when storing a document. You added a timestamp which is useful, but it might be problematic when sorting messages. The messages should be presented in the order they were emitted by a client. 

As timestamps are generated from the pilot itself (as logs are printed out), if we sort them by timestamps, we could add the right numbering without losing space?

PR #552, opened 2025-05-26
fix: joserfc breaking changes

aldbr:
Aims at fixing https://github.com/DIRACGrid/diracx/actions/runs/15248175086/job/42878643647?pr=421#step:12:1222

### DIFF
--- a/diracx-routers/src/diracx/routers/auth/token.py
+++ b/diracx-routers/src/diracx/routers/auth/token.py
@@ -11,7 +11,6 @@ from joserfc.errors import DecodeError, ExpiredTokenError, InvalidKeyIdError
 
 from diracx.core.exceptions import (
     DiracHttpResponseError,
-    ExpiredFlowError,
     InvalidCredentialsError,
     PendingAuthorizationError,
 )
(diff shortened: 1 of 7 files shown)


diracx-routers/src/diracx/routers/auth/token.py
```diff
@@ -158,7 +157,7 @@ async def get_oidc_token(
             detail=str(e),
             headers={"WWW-Authenticate": "Bearer"},
         ) from e
-    except JoseError as e:
+    except (DecodeError, ExpiredTokenError) as e:
```
fstagni, line 160: I feel like these should be treated differently, e.g. 

```suggestion
    except (ExpiredTokenError) as e:
        raise HTTPException(
            status_code=status.HTTP_401_UNAUTHORIZED,
            detail=f"Expired JWT: {e}",
        ) from e
    except (DecodeError) as e:
        raise HTTPException(
            status_code=status.HTTP_400_BAD_REQUEST,
            detail=f"Invalid JWT: {e}",
        ) from e
```

fstagni (approved):

PR #556, opened 2025-06-02
Postgres with the SmarterDateTime

AcquaDiGiorgio:
This PR is a rebase of https://github.com/DIRACGrid/diracx/pull/297 onto https://github.com/DIRACGrid/diracx/pull/454 to test @ryuwd 's changes with PostgreSQL.

The main changes are:

First, the dialect name of the naive dictionary was changed from "postgres" to "postgresql". A way to remove this hard-coded dialect string is by using the sqlalchemy dialect classes:
```python
from sqlalquemy.dialects import mysql, postgresql, oracle, sqlite

mysql_dialect = mysql.dialect().name
...
```

Second, the partial class `DateNowColumn` now implements `SmarterDateTime`.

And finally, which was the trickiest change, relates to the table creation and timezone aware insertions.

During table creation, the timezone requirement depends of the value of the timezone passed to the `DateTime` class in the `impl` parameter of the `SmarterDateTime` class.
Mysql creates the same creation statement independent of the value, however Postgres does not.

If the `DateTime` is created with `timezone=True`, the statement is created with the keywords `WITH TIME ZONE`:
```sql
CREATE TALBE "X" (
	"Y" TIMESTAMP 
		WITH TIME ZONE 
		DEFAULT TIMEZONE('utc', CURRENT_TIMESTAMP)
		NOT NULL,
)
```
If it created without passing it a value (default to False), it instead creates the columns with the keywords `WITHOUT TIME ZONE`:
```sql
CREATE TALBE "X" (
	"Y" TIMESTAMP 
		WITHOUT TIME ZONE 
		DEFAULT TIMEZONE('utc', CURRENT_TIMESTAMP) 
		NOT NULL,
)
```

The issue here is that if we insert a datetime that contains a timezone to the second table, it will throw an exception saying that the types mismatch and stop the query.

There are two ways of solving this:

1) Setting the timezone as True in the DateTime's impl parameter of the SmarterDateTime and maintaining postgres as timezone aware. 
2) Not setting the timezone and setting PostgreSQL as naive.

The second option is in my opinion the more sensible solution, because it forces the same behaviour to all dialects. This is what I ended up doing.


fstagni:
After discussions in https://indico.cern.ch/event/1550175/ ( -- see its minutes) we sadly decided that we can't support postgreSQL in a neat way. The main reason for giving up PostgreSQL for the time being as we would have too many things to do in DIRAC’s MySQL for fixing how the dates are stored.

So, closing :-(

PR #570, opened 2025-06-13
feat: add pilot database and router

### LINKED ISSUE #417: Open access and require auth not working at all

### LINKED ISSUE #960: Pilot Manager service and DB in DiracX
In this issue we just keep track of #570 where the work is already done.

Robin-Van-de-Merghel:
Split of #421 , first part : pilot management

closes #960


diracx-db/src/diracx/db/sql/utils/base.py
```diff
@@ -227,6 +227,47 @@ async def ping(self):
...
+        model: Any,
+        parameters: list[str] | None,
+        search: list[SearchSpec],
+        sorts: list[SortSpec],
+        *,
+        distinct: bool = False,
+        per_page: int = 100,
+        page: int | None = None,
+    ) -> tuple[int, list[dict[Any, Any]]]:
+        """Search for pilots in the database."""
```
fstagni, line 241:
```suggestion
        """Search in a SQL database, with filters"""
```

diracx-core/src/diracx/core/exceptions.py
```diff
@@ -100,3 +101,43 @@ def __init__(self, job_id, detail: str | None = None):
 
 class NotReadyError(DiracError):
     """Tried to access a value which is asynchronously loaded but not yet available."""
+
+
+class DiracFormattedError(DiracError):
```
fstagni, line 106: All exceptions apart from the 2 below inherit from `DiracError`, and this one is further customization for only a few related to pilot. What if instead you modify `DiracError`?

diracx-db/src/diracx/db/sql/job/db.py
```diff
@@ -69,7 +57,7 @@ async def summary(self, group_by, search) -> list[dict[str, str | int]]:
             if row.count > 0  # type: ignore
         ]
 
-    async def search(
+    async def search_jobs(
```
fstagni, line 60: I'd leave the name `search` here, and use `super` below.

fstagni (commented):
First review.

diracx-db/src/diracx/db/sql/job/db.py
Robin-Van-de-Merghel: We won't search for other things in jobs later ? (job parameters, job commands, etc.)

diracx-db/src/diracx/db/sql/pilots/db.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        stmt = delete(PilotAgents).where(PilotAgents.pilot_stamp.in_(pilot_stamps))
+
+        res = await self.conn.execute(stmt)
+
+        if res.rowcount != len(pilot_stamps):
+            raise PilotNotFoundError(data={"pilot_stamps": str(pilot_stamps)})
+
+    async def clear_pilots_bulk(
+        self, cutoff_date: datetime, delete_only_aborted: bool
```
fstagni, line 132:
```suggestion
        self, cutoff_date: datetime, delete_only_aborted: bool = False
```

diracx-core/src/diracx/core/models.py
```diff
@@ -272,3 +272,18 @@ class JobCommand(BaseModel):
     job_id: int
     command: Literal["Kill"]
     arguments: str | None = None
+
```
fstagni, line 275: You should add `PilotStatus` similar to what is done for `JobStatus`, and use later on.

diracx-db/src/diracx/db/sql/pilots/db.py
```diff
@@ -0,0 +1,247 @@
...
+    ) -> int:
+        """Bulk delete pilots that have SubmissionTime before the 'cutoff_date'.
+        Returns the number of deletion.
+        """
+        # TODO: Add test (Millisec?)
+        stmt = delete(PilotAgents).where(PilotAgents.submission_time < cutoff_date)
+
+        # If delete_only_aborted is True, add the condition for 'Status' being 'Aborted'
+        if delete_only_aborted:
+            stmt = stmt.where(PilotAgents.status == "Aborted")
```
fstagni, line 142: You should get the status from model.py (see comment in it).

diracx-db/src/diracx/db/sql/pilots/db.py
```diff
@@ -0,0 +1,247 @@
...
+            }
+            for stamp in pilot_stamps
+        ]
+
+        # Insert multiple rows in a single execute call and use 'returning' to get primary keys
+        stmt = insert(PilotAgents).values(values)  # Assuming 'id' is the primary key
+
+        await self.conn.execute(stmt)
+
+    async def associate_pilot_with_jobs(
```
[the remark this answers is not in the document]
Robin-Van-de-Merghel: We only associate one at a time ?

diracx-db/src/diracx/db/sql/pilots/db.py
```diff
@@ -0,0 +1,247 @@
...
+        # Prepare the list of dictionaries for bulk insertion
+        values = [
+            {
+                "PilotJobReference": pilot_references.get(stamp, stamp),
+                "VO": vo,
+                "GridType": grid_type,
+                "SubmissionTime": now,
+                "LastUpdateTime": now,
+                "Status": "Submitted",
+                "PilotStamp": stamp,
```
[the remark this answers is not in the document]
fstagni: as common pars

diracx-routers/src/diracx/routers/pilots/management.py
```diff
@@ -0,0 +1,232 @@
...
+    try:
+        await delete_pilots_by_stamps_bulk(pilot_db=pilot_db, pilot_stamps=pilot_stamps)
+    except PilotNotFoundError as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="At least one pilot has not been found.",
+        ) from e
+
+
+@router.delete("/management/pilot/interval", status_code=HTTPStatus.NO_CONTENT)
```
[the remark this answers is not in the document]
Robin-Van-de-Merghel: If I do so, should I have a parameter similar to `GrantType` : `ClearType`

> `ClearType.BY_STAMP`, `ClearType.BY_AGE`

Or else I would deal with `if` statements and would have to prioritize a method over the other

diracx-routers/src/diracx/routers/pilots/query.py
```diff
@@ -0,0 +1,149 @@
...
+                        "BenchMark": 63.1,
+                    },
+                ]
+            }
+        },
+    },
+}
+
+
+@router.post("/management/search", responses=EXAMPLE_RESPONSES)
```
Robin-Van-de-Merghel, line 107: To be changed

diracx-logic/src/diracx/logic/pilots/management.py
```diff
@@ -0,0 +1,100 @@
...
+        vo=vo,
+        grid_type=grid_type,
+        grid_site=grid_site,
+        destination_site=destination_site,
+        pilot_references=pilot_job_references,
+        status_reason=status_reason,
+    )
+
+
+async def clear_pilots(
```
Robin-Van-de-Merghel, line 49: [Needs work]: https://github.com/DIRACGrid/DIRAC/blob/1ea0d9a33e6f62ebd46673b6e5e060448ad887d0/src/DIRAC/WorkloadManagementSystem/DB/PilotAgentsDB.py#L191-L210

Robin-Van-de-Merghel: (For `deletePilots`: delete mapping also, same later for logs #511 + `PilotOutput`)

Robin-Van-de-Merghel: Left open as a reminder

diracx-routers/src/diracx/routers/pilots/management.py
```diff
@@ -0,0 +1,252 @@
...
+            vo=vo,
+            grid_type=grid_type,
+            grid_site=grid_site,
+            destination_site=destination_site,
+            pilot_job_references=pilot_references,
+            status_reason=status_reason,
+        )
+
+        # Logs credentials creation
+        logger.debug(
```
Robin-Van-de-Merghel, line 89: Still useful?

diracx-logic/src/diracx/logic/pilots/management.py
```diff
@@ -0,0 +1,117 @@
...
+    pilot_db: PilotAgentsDB,
+    pilot_stamps: list[str],
+    vo: str,
+    grid_type: str,
+    grid_site: str,
+    destination_site: str,
+    status: str,
+    pilot_job_references: dict[str, str] | None,
+):
+    # [IMPORTANT] Check unicity of pilot references
```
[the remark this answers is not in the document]
Robin-Van-de-Merghel:
```suggestion
    # [IMPORTANT] Check unicity of pilot stamps
```

fstagni (approved):

diracx-logic/src/diracx/logic/pilots/management.py
```diff
@@ -0,0 +1,120 @@
...
+    status: str,
+    pilot_job_references: dict[str, str] | None,
+):
+    # [IMPORTANT] Check unicity of pilot stamps
+    # If a pilot already exists, we raise an error (transaction will rollback)
+    existing_pilots = await get_pilots_by_stamp(
+        pilot_db=pilot_db, pilot_stamps=pilot_stamps
+    )
+
+    # If we found pilots from the list, this means some pilots already exists
```
Robin-Van-de-Merghel, line 33:
```suggestion
    # If we found pilots from the list, this means some pilots already exist
```
?

diracx-routers/src/diracx/routers/pilots/management.py
```diff
@@ -0,0 +1,259 @@
...
+        vo_constraint=vo_constraint,
+    )
+
+
+EXAMPLE_UPDATE_FIELDS = {
+    "Update the BenchMark field": {
+        "summary": "Update BenchMark",
+        "description": "Update only the BenchMark for one pilot.",
+        "value": {
+            "pilot_stamps_to_fields_mapping": [
```
Robin-Van-de-Merghel, line 164: Reduce?

diracx-db/src/diracx/db/sql/utils/base.py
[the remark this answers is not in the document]
Robin-Van-de-Merghel: #608

aldbr:
It's ready to be reviewed again @fstagni 

### `diracx-core`

- **`PilotFieldsMapping` → `PilotMetadata`.** Renamed for clarity ("fields mapping" didn't say what it was), and every field now has a Pydantic `Field(description=...)` so the OpenAPI schema is self-documenting. 
- **`ScalarSearchSpec.value`: removed `datetime`.** Wasn't needed.

### `diracx-db`

- **Method rename pass:** `add_pilots` → `register_pilots`, `add_jobs_to_pilot` → `assign_jobs_to_pilot`, `update_pilot_fields` → `update_pilot_metadata`. Names now match what they do.
- **Heterogeneous bulk metadata update.** `update_pilot_metadata` accepts a list of `PilotMetadata` objects where each row may set a different subset of fields. Mirrors the pattern used in `JobDB.set_job_attributes`.
- **Removed unused `search_pilot_to_job_mapping`** (no callers).
- **Schema typing cleanup.** `Optional[datetime]` → `datetime | None`, dropped the `typing.Optional` import to match the rest of `diracx-db/sql`.

### `diracx-logic`

- **`logic/pilots/management.py`.** `register_new_pilots`, `delete_pilots`, `update_pilots_metadata`, `assign_jobs_to_pilot`. The `delete_pilots` entry point handles both stamp-based and age-based branches; the age-based branch is reserved for the maintenance task worker, am I understanding correctly?
- **`logic/pilots/query.py`.** `search`, `summary`, `get_pilots_by_stamp`, plus the `JobID` pseudo-parameter rewriter (see below). Adds a `_add_vo_constraint` helper that mirrors the intra-VO filter pattern from `logic/jobs/query.py`.
- **`JobID` ↔ `PilotStamp` pseudo-parameters.** Both `POST /api/jobs/search` and `POST /api/pilots/search` now accept a pseudo-parameter that is transparently resolved through `JobToPilotMapping`:
  - `POST /api/pilots/search` accepts `JobID` (`eq`/`in` only), rewritten into a `PilotID IN (...)` filter.
  - `POST /api/jobs/search` accepts `PilotStamp` (`eq`/`in` only), rewritten into a `JobID IN (...)` filter.
  - Mixing a pseudo-parameter with its real counterpart in the same request body is refused with `InvalidQueryError`. Unsupported operators are rejected with the same error (the join semantics would be ambiguous). If the resolution produces an empty set, the search short-circuits to an empty result.

### `diracx-routers`

- **`POST /api/pilots/`** — bulk register. Rejects any caller with `GENERIC_PILOT` property sending more than one stamp per call (limits the blast radius of a stolen pilot credential). Non-admins can only register pilots for their own VO.
- **`DELETE /api/pilots/`** — stamp-based delete only. Deletes the pilot rows, their logs, and their job associations. Age-based retention is *not* exposed on the HTTP API (handled by a diracx-task?).

### Docs

Updated `docs/admin/explanations/pilots.md` and `docs/dev/explanations/pilots.md`.

fstagni:
Great, but now there are conflicts, and one test is not passing. Can you fix?

aldbr:
> Great, but now there are conflicts, and one test is not passing. Can you fix?

Yes it's fixed now

aldbr:
`pytest-integration` not passing, should be fixed with https://github.com/DIRACGrid/diracx/pull/924 :crossed_fingers:

aldbr:
> ddev: 
> part of a bigger plan with 3 axis: pilot management (user), pilot security (pilot tokens), pilot transition in DIRAC
> only missing piece is the getPilotOutput: could be handled with a DIRAC Agent to fetch the pilot output and put it in the DB for now
> also the current DIRAC Webapp does not communicate to diracx 

> There was a plan (may be not written anywhere?):
> - pilot can talk to DIRAC and DiracX
> - DiracX can interact with pilot
> Either we revisit that plan or we write it down in an ADR

> @fstagni will double check and see if getPilotOutput is the only issue. If it's the case, we just isolate it in DIRAC (outside of the PilotManagerHandler) and we move on with the legacy adaptor. Will closely be followed by issues if any problem occurs afterwards.

diracx-routers/src/diracx/routers/pilots/management.py
```diff
@@ -0,0 +1,172 @@
...
+from .access_policies import (
+    ActionType,
+    CheckPilotManagementPolicyCallable,
+)
+
+router = DiracxRouter()
+
+
+@router.post("/")
+async def register_pilots(
```
chrisburr: I don't think this needs to support bulk registration? The SiteDirectors in DIRAC should be accessing the DB directly and only one pilot will appear from the vaccum at a time?

diracx-routers/src/diracx/routers/pilots/management.py
```diff
@@ -0,0 +1,172 @@
...
+        grid_type=grid_type,
+        grid_site=grid_site,
+        destination_site=destination_site,
+        pilot_job_references=pilot_references,
+        status=pilot_status,
+    )
+
+
+@router.delete("/", status_code=HTTPStatus.NO_CONTENT)
+async def delete_pilots(
```
chrisburr: I think this should only be called via diracx tasts and DIRAC agents which are accessing the DB directly?

aldbr: I agree, but IIRC this is part of the `PilotManagerHandler` and I guess it's here because of that.
@fstagni can you double check please? Because if it's only used by the webapp / CLI (which I assume is the case), then we can move it to the web utilities for DIRAC as `getPilotOutput`.

fstagni: The corresponding DIRAC RPC call was actually removed by https://github.com/DIRACGrid/DIRAC/pull/8624. So, will remove also from here.

aldbr (approved):

PR #580, opened 2025-07-01
Alembic triggers

AcquaDiGiorgio:
With this PR I want to display the possibility to create and drop triggers using Alembic. 

I left some code that achieves the same thing using just SQLAlchemy commented to have something to compare them, but they must not work along side each other.

# Known Issues

- Currently MySQL specific. PostgreSQL is the only language to have a [special package](https://github.com/olirice/alembic_utils) to deal with this issues. For other dialects, including MySQL, it doesn't seem to be anything similar and it has to be done this way.

- Auto-generation of certain DiracX specific data types (not the Triggers) have some problems. I have seen this behaviour with the enums, which it doesn't know how the init function works.

- The url for the connection must be from a privileged user, so we cannot use the environment variable obtained from `BaseSQLDB.available_urls()["DB_NAME"]`.

- The Alembic directory structure of the migrations has been initialized only for the `JobDB` database, but it can be either copied to other databases, or we can try to have a singular alembic script and multiple migrations. Apparently is not easy or recommended, but there are guides, such as [this one](https://learningtotest.com/2021/06/17/managing-alembic-migrations-with-a-single-alembic-ini-env-py/), that makes it look simple.

# Recommended documentation

[Alembic Plugins](https://alembic.sqlalchemy.org/en/latest/api/operations.html#operation-plugins): Explains how to create new operations for not implemented SQL statements such as `CREATE TRIGGER`. 

[Alembic Autogeneration](https://alembic.sqlalchemy.org/en/latest/api/autogenerate.html#autogenerating-custom-operation-directives): Shows how to make Alembic know how to generate the code of the plugins.

[GitHub Issue - PG Triggers](https://github.com/sqlalchemy/alembic/issues/686): Someone had a similar issue with PostgreSQL and the comments illustrate how to modify alembic's context to automatically generate code using the directives in the `.mako` template file


diracx-db/src/diracx/db/sql/job/migrations/versions/79fce7dece6a_init.py
fstagni, line 1: For my understanding, is this file created by alembic?

diracx-db/src/diracx/db/sql/job/migrations/versions/79fce7dece6a_init.py
```diff
@@ -0,0 +1,184 @@
...
+        sa.Column("JobID", sa.Integer(), nullable=False),
+        sa.Column("Name", sa.String(length=100), nullable=False),
+        sa.Column("Value", sa.Text(), nullable=False),
+        sa.ForeignKeyConstraint(["JobID"], ["Jobs.JobID"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("JobID", "Name"),
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
```
fstagni, line 162: This seems quite ... wrong!

AcquaDiGiorgio: What part in particular? The table dropping?
If that's the case, remember that this is the very first version of the database and the previous state is an empty one, so it drops everything it created.

diracx-db/src/diracx/db/sql/job/migrations/versions/79fce7dece6a_init.py
fstagni: Can they stay in a specific directory, e.g. `diracx-db/src/diracx/db/sql/_generated` ?

diracx-db/src/diracx/db/sql/job/migrations/versions/79fce7dece6a_init.py
fstagni: Ah that's certainly not what we want. The version we have is "inherited" from DIRAC so no dropping!

diracx-db/src/diracx/db/sql/job/migrations/versions/79fce7dece6a_init.py
AcquaDiGiorgio: Yes! There are 2 configuration variables in the `alembic.ini` file called `script_location` and `version_locations` that set up both the location of the Alembic environment and its revisions.
 
- `script_location` defaults to `%(here)s/migrations`, where `%(here)s` is the location of the `.ini`. 
- `version_locations` defaults to `<script_location>/versions` but it can also be configured.

diracx-db/src/diracx/db/sql/job/migrations/versions/79fce7dece6a_init.py
AcquaDiGiorgio: So we will never downgrade? And for upgrades, if a table is removed from the schema, does it need to stay in the DB?

fstagni: We might want to downgrade at some point (or at least having the possibility to do so).

For the upgrades: if a table is removed from the schema it does not need to stay in the DB.

fstagni: I think so.

aldbr:
I am closing this PR as:
- it would take some time to rebase
- we first need to have an ADR describing the scope of Alembic and how we expect to use it

Feel free to rebase and reopen once the ADR is in place and the issue is discussed in a ddev meeting

Cheers

PR #593, opened 2025-07-15
Add job monitoring utilities to the routers

Robin-Van-de-Merghel:
To update the job monitoring, we need a `getInputData` and a `get_job_parameters` router.


diracx-db/src/diracx/db/os/utils.py
```diff
@@ -198,12 +198,24 @@ async def upsert(self, vo: str, doc_id: int, document: Any) -> None:
...
 
     async def search(
-        self, parameters, search, sorts, *, per_page: int = 100, page: int | None = None
-    ) -> list[dict[str, Any]]:
+        self,
+        parameters,
+        search,
+        sorts,
+        *,
+        per_page: int = 10000,
```
fstagni, line 206: Seems like a very large number, why?

diracx-db/src/diracx/db/sql/dummy/db.py
```diff
@@ -21,7 +21,7 @@ class DummyDB(BaseSQLDB):
     # This needs to be here for the BaseSQLDB to create the engine
     metadata = DummyDBBase.metadata
 
-    async def summary(self, group_by, search) -> list[dict[str, str | int]]:
+    async def dummy_summary(self, group_by, search) -> list[dict[str, str | int]]:
```
fstagni, line 24: why?

Robin-Van-de-Merghel: I rewrote the `search` function for `BaseDB`. The new `search` function has to be generic, and takes all regular parameters plus a `Model` which is for example `PilotAgents` or `PilotSecrets`.

Calling directly `db.summary` will require this `Model` parameter, whereas `dummy_summary` will add this `Model` parameter.

Maybe I can rename the generic `search` function : `_search` (to make it private)

diracx-routers/src/diracx/routers/jobs/query.py
```diff
@@ -178,12 +182,63 @@ async def search(
...
+    job_db: JobDB,
+    check_permissions: CheckWMSPolicyCallable,
+) -> list[dict[str, Any]]:
+    """Fetches a job's input data."""
+    await check_permissions(action=ActionType.MANAGE, job_db=job_db, job_ids=[job_id])
+
+    return await get_input_data_bl(job_db=job_db, job_id=job_id)
+
+
+@router.get("/{job_id}/parameters")
```
Robin-Van-de-Merghel, line 197: To be discussed.

Robin-Van-de-Merghel:
See comments above, we need to review it...

diracx-routers/src/diracx/routers/jobs/query.py
fstagni: Fine as it is, I think.

diracx-routers/src/diracx/routers/jobs/query.py
```diff
@@ -178,12 +182,63 @@ async def search(
...
+) -> list[dict[str, Any]]:
+    """Get job parameters."""
+    await check_permissions(action=ActionType.MANAGE, job_db=job_db, job_ids=[job_id])
+
+    return await get_job_parameters_bl(
+        job_parameters_db=job_parameters_db, job_id=job_id
+    )
+
+
+@router.get("/{job_id}/jdl")
```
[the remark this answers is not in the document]
fstagni: Fine as it is, I think

diracx-routers/src/diracx/routers/jobs/query.py
aldbr: Don't we want to handle job attributes and parameters as "metadata" in DiracX?

PR #605, opened 2025-07-23
New operators for job db

TheauW:
I've added two new operators for the job search. These operators come with tests.

### DIFF
--- a/diracx-db/tests/jobs/test_job_db.py
+++ b/diracx-db/tests/jobs/test_job_db.py
@@ -218,14 +218,23 @@ async def test_search_conditions(populated_job_db):
         assert result
         assert len(result) == 100
 
-        # Search for a specific scalar condition: JobID regex '\d'
+        # Search for a specific scalar condition: JobID regex 'owner[0-3]+'
+        # owner0, owner1, owner2, owner3 (4 jobs)
+        # owner11 -> owner39 (30 jobs)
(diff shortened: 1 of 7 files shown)


diracx-db/tests/jobs/test_job_db.py
Robin-Van-de-Merghel, line 1: With:

```py
        # Search for a bad regex
        condition = ScalarSearchSpec(
            parameter="Owner", operator=ScalarSearchOperator.REGEX, value="owner[0-9+"
        )
        total, result = await job_db.search([], [condition], [])
        assert total == 100
        assert result
        assert len(result) == 100
```

You get an `OperationalError` instead of an `InvalidQueryError`, which will lead in a 500 error in the routers -- which is not intended.

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) user-defined function raised exception
E               [SQL: SELECT count(*) AS count_1 
E               FROM (SELECT "Jobs"."JobID" ...
E               FROM "Jobs" 
E               WHERE "Jobs"."Owner" REGEXP ?) AS anon_1]
E               [parameters: ('owner[0-9+',)]
E               (Background on this error at: https://sqlalche.me/e/20/e3q8)
```

I see two possibilities:

1. Add a `try/catch` to catch an operational error and raise InvalidQuery (generic)
2. Add a verification for regex

PR #606, opened 2025-07-29
fix: check AUTOREST_HOME directory is created if it doesn't exist

ryuwd:
Fixes a bug on macOS where the directory would not be created causing the client generation to fail.


chrisburr:
We set `AUTOREST_HOME` in the pixi.toml so it isn't in the user's home directory (which causes some weird reproducibility issues over time).

chrisburr (approved):

PR #608, opened 2025-08-04
refactor: Made search and summary generic to reuse it easily

Robin-Van-de-Merghel:
Extracted from #570


diracx-db/src/diracx/db/sql/utils/base.py
```diff
@@ -227,6 +231,71 @@ async def ping(self):
...
+        self,
+        table: Any,
+        parameters: list[str] | None,
+        search: list[SearchSpec],
+        sorts: list[SortSpec],
+        *,
+        distinct: bool = False,
+        per_page: int = 100,
+        page: int | None = None,
+    ) -> tuple[int, list[dict[Any, Any]]]:
```
chrisburr, line 244:
```suggestion
    ) -> tuple[int, list[dict[str, Any]]]:
```

diracx-db/src/diracx/db/sql/utils/base.py
```diff
@@ -227,6 +231,71 @@ async def ping(self):
         except OperationalError as e:
             raise SQLDBUnavailableError("Cannot ping the DB") from e
 
+    async def _search(
+        self,
+        table: Any,
```
chrisburr, line 236: Maybe use `reveal_type(...)` to figure out what the correct type hint is here

chrisburr (approved):

PR #637, opened 2025-08-22
chore(ci): pixi --no-install does not work anymore in the CI

aldbr:


.github/workflows/main.yml
```diff
@@ -23,15 +23,6 @@ jobs:
         with:
           run-install: false
           post-cleanup: false
-      - name: Apply workarounds
-        run: |
-          # Workaround for https://github.com/prefix-dev/pixi/issues/3762
-          sed -i.bak 's@editable = true@editable = false@g' pixi.toml
```
chrisburr, line 29: We should keep this line

PR #669, opened 2025-09-24
feat: add package versions to well-known installation metadata endpoint

ryuwd:
Add package version metadata to the `/.well-known/dirac-metadata` endpoint and also for gubbins

### DIFF
--- a/diracx-logic/src/diracx/logic/auth/well_known.py
+++ b/diracx-logic/src/diracx/logic/auth/well_known.py
@@ -1,8 +1,27 @@
 from __future__ import annotations
 
+from importlib.metadata import PackageNotFoundError
+
+import pkg_resources
+
 from diracx.core.config.schema import Config
+from diracx.core.extensions import extensions_by_priority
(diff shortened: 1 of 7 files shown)


diracx-logic/src/diracx/logic/auth/well_known.py
```diff
@@ -1,10 +1,29 @@
...
 from diracx.core.models import GroupInfo, Metadata, OpenIDConfiguration
 from diracx.core.settings import AuthSettings
 
 
+def get_package_version(extension_name: str) -> dict[str, str]:
+    """Get package versions for a given diracx extension."""
+    installed_packages = pkg_resources.working_set
+    matching_packages = {}
+    for package in installed_packages:
+        if package.project_name.startswith(f"{extension_name}-"):
```
chrisburr, line 18: It would be better not to rely on this convention, maybe we walk the entrypoints?

ryuwd:
This is already done in `openapi.json`

PR #675, opened 2025-10-10
partition authdb

chaen:


chaen:
if you have 2 billion rows like LHCb, you want to do it in a bit smarter way than just running the alter schema command. Fortunately, I think it is not the case of anyone :-) 
The `init-sql` job prints that in its output.

PR #690, opened 2025-10-27
fix: support additional config hints

aldbr:
Aims at solving an issue we get in the certification environments, making `legacy_adaptor` a bit more robust, and adding tests around it.


aldbr:
Note: real fix is https://github.com/DIRACGrid/cs_store_dirac_certif/pull/3

The main benefit from this PR is the unit test I think.

We should of course discuss about a more robust solution to avoid breaking the yaml config each time we add a new field in the CS. Adding `extra=ignore` in all the sections we think aren't complete is probably a good solution as suggested by @chrisburr

chrisburr (approved):

PR #715, opened 2025-12-03
Chore: following suggestions from repo-review

fstagni:
From https://learn.scientific-python.org/development/guides/repo-review/?repo=DIRACGrid%2Fdiracx&ref=main&refType=branch

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,7 @@ ignore-words-list = ["CheckIn", "dependant"]
 
 [tool.mypy]
 python_version = "3.11"
+strict = false
 files = [
   "diracx-api/src/**/*.py",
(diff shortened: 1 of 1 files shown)


diracx-client/src/diracx/client/patches/jobs/common.py
```diff
@@ -14,8 +14,8 @@
 
 import json
 from io import BytesIO, IOBase
-from typing import Any, IO, Dict, TypedDict, Union, Unpack, cast, Literal
-
+from typing import Any, IO, Dict, Union, Unpack, cast, Literal
+from typing_extensions import TypedDict
```
aldbr: Not sure to understand this change (`TypedDict` has been introduced from python 3.8)

PR #721, opened 2025-12-18
Table extensions examples

fstagni:


extensions/gubbins/gubbins-db/src/gubbins/db/sql/jobs/schema.py
```diff
@@ -1,17 +1,134 @@
...
+    Index,
     Integer,
     String,
+    Text,
 )
 
+# NOTE: We need to remove the original Jobs table from metadata because
+# GubbinsJobs modifies an existing column (job_name size from 128 to 512).
+# Simple inheritance works fine for ADDING columns (see GubbinsInputData below),
+# but MODIFYING existing columns in SQLAlchemy requires replacing the table definition.
```
chrisburr: > MODIFYING existing columns

IIRC we had said that we didn't want to support this, do you have a use case in mind?

aldbr: > We explicitly said we would not do that because it has too many implications. We will even have to modify some LHCb database to follow that.

Is that written anywhere in the documentation?
I found that block of documentation about DB extensions, but it's not explicitly mentioned: https://diracx.diracgrid.org/en/latest/dev/explanations/extensions/#extended-db

If not written anywhere, I think we should clearly mention it there (this section of the documentation is probably the right place)

(In the future, as we discussed, this should likely go through a Github Issue for discussion first)

chrisburr:
Closing this until we have a use case to consider.

chrisburr (changes requested):
This PR is mixing pyproject.toml changes with DB changes and also needs to have the column modification example removed.

aldbr:
Just wondering, do we even give the possibility of extending DBs by adding columns?
Is there any use case? From what I understand, only LHCb is concerned.

chaen:
I think there are enough cases here that the docs can be explicit with:
* what is not in gubbins is NOT supported
* special mention of "adding a column is NOT supported because x,y,z"

Would you all agree ?

PR #728, opened 2026-01-19
docs: document the release process

chaen:


docs/dev/reference/making-releases.md
```diff
@@ -0,0 +1,72 @@
...
+The `diracx-web` version is only very loosely coupled with the `diracx` version.
+A new `diracx-web` release will trigger a new `diracx-charts` updating:
+
+- the chart version
+- the container versions
+- NOT the AppVersion
+
+## diracx release
+
+Releases are managed by `ReleasePlease`. The token used to perform it belongs to `diracbot` github user
```
aldbr: Shall we document `diracx-web` release here? I think yes.
It would be very similar to `diracx` release.
The only differences are:
- we are not using `diracbot` because we don't need it (at least for now)
- `pypi release` becomes `npm release` for `diracx-web-components`. See https://www.npmjs.com/package/@dirac-grid/diracx-web-components

docs/dev/reference/making-releases.md
```diff
@@ -0,0 +1,72 @@
...
+- new [docker images](https://github.com/DIRACGrid/diracx/)
+
+### Force version number
+
+You can force a specific version number following the [doc](https://github.com/googleapis/release-please?tab=readme-ov-file#how-do-i-change-the-version-number)
+
+```bash
+git commit --allow-empty -m "chore: release 0.0.3" -m "Release-As: v0.0.3"
+git push upstream HEAD:main
+```
```
aldbr: This is also written in `make-a-release.md`. Shouldn't we add a reference to it here?

docs/dev/how-to/make-a-release.md
```diff
@@ -0,0 +1,46 @@
+This page describes how to mechanically make releases of DiracX and extensions. See the [reference](../reference/making-releases.md) documentation for more details about how releases are made.
+
+# How to make a new DiracX release
+
+1. Merge any PR you want in your release
```
fstagni: Remove this step, it's done asynchronously.

Issue #612, opened 2025-08-05
Use the new-style sqlalchemy DeclarativeBase


chrisburr:
Actually we should use https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#migrating-an-existing-mapping so we can type this as Declarative base.

For another PR.

_Originally posted by @chrisburr in https://github.com/DIRACGrid/diracx/pull/608#discussion_r2253369866_

Issue #433, opened 2025-03-31
SQLAlchemy typing and drop deprecated mypy plugin


chaen:
we currently have `#type: ignore` comments in places like that:

https://github.com/DIRACGrid/diracx/blob/a37ff0db0dab93ce90d44938310ba35ed1f1cbd9/diracx-db/src/diracx/db/sql/auth/db.py#L97

This is because we need to specify the types at the table level 

https://github.com/DIRACGrid/diracx/blob/a37ff0db0dab93ce90d44938310ba35ed1f1cbd9/diracx-db/src/diracx/db/sql/auth/schema.py#L44

Should be changed to something like 

```python
user_code: Mapped[str] = mapped_column("UserCode", String(USER_CODE_LENGTH), primary_key=True)
```

Conversion guide:

https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#new-typing-support-in-core-and-orm-stubs-extensions-no-longer-used

Also the mypy plugin is no longer supported and should be dropped: https://github.com/DIRACGrid/diracx/issues/438

PR #748, opened 2026-01-22
Refactor: Change the DeclarativeBase sqlalchemy style

HeloiseJoffe:
closed: #612 

Changes: 
- Replaced declarative_base with DeclarativeBase style
- Cast '__table__' to Table type to satisfy mypy

Questions: 
There is other new styles in SQLAlchemy, for example the declarative use of Column is replaced by mapped_column.
Should this be applied as well? 
In the Bookkeeping schema, this has been done, but the other changes haven’t been applied. Is there a reason for that?

### DIFF
--- a/diracx-db/src/diracx/db/sql/auth/db.py
+++ b/diracx-db/src/diracx/db/sql/auth/db.py
@@ -163,7 +163,7 @@ class AuthDB(BaseSQLDB):
         for _ in range(MAX_RETRY):
             user_code = "".join(
                 secrets.choice(USER_CODE_ALPHABET)
-                for _ in range(DeviceFlows.user_code.type.length)  # type: ignore
+                for _ in range(DeviceFlows.user_code.type.length)
             )
             device_code = secrets.token_urlsafe()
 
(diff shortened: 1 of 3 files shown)


ryuwd (commented):
Thank you for the hard work! I'm happy with how you've migrated the tables, there are just a couple more old utility functions we should try to get rid of if possible (see comments)

A TODO for the future - we should keep track with tests and test fixtures how these schema definitions change the DB schema expected by SQLAlchemy

aldbr (commented):
I just have one last minor comment. The issue was initially opened for that specific comment: https://github.com/DIRACGrid/diracx/pull/608#discussion_r2253369866

Basically to change `Any` in `search` and `summary` I guess:
https://github.com/DIRACGrid/diracx/blob/faf6b44c4954756210fc6d0a806422051ba04249/diracx-db/src/diracx/db/sql/utils/base.py#L253

https://github.com/DIRACGrid/diracx/blob/faf6b44c4954756210fc6d0a806422051ba04249/diracx-db/src/diracx/db/sql/utils/base.py#L293

Do you think we can have something better now?

aldbr:
Very last comment, sorry :sweat_smile: 
While I was going through the open issues, I noticed that https://github.com/DIRACGrid/diracx/issues/433 was very similar but contains a few additional info about `type ignore` and `mypy`.
Could you check that please?

aldbr (approved):

PR #766, opened 2026-02-05
feat: check for __all__ presence in __init__.py files

### LINKED ISSUE #426: Make sure `__all__` is present in the `__init__` files and up to date
Based on https://github.com/DIRACGrid/diracx/issues/406#issuecomment-2706467195

@fstagni noticed that `JobParametersDB` was not listed in `__all__` within `diracx.routers.dependencies`. It means that we are forgetting to add new variables/functions in the public API over time, which might possibly create inconsistencies.

- If there is an automatic rule (`mypy`?) checking that imported variables and functions are all part of `__all__` if it exists, then we should probably go for it (though I haven't found anything by doing a quick search)
- Else we should at least provide a piece of documentation about that for the developers (`CODING_CONVENTION.md`)

AcquaDiGiorgio:
See #426

I tried something more sofisticated in python using `importlib` to check the variables in the module, both instantiated and imported, but there were multiple situations that were extremely difficult to check.

We also might want to set a common way of defining this variable, using either tuples or lists:
- **diracx/routers/health/\_\_init\_\_.py:** `__all__ = ["router"]`
- **diracx/db/os/\_\_init\_\_.py** `__all__ = ("JobParametersDB",)`


chrisburr (changes requested):

.github/workflows/check_init_files_precommit_hook.sh
aldbr: Any specific reason for making a bash script instead of a pixi task as done here: https://github.com/DIRACGrid/diracx/blob/9cc792e4849190f7ec01b2cc0f2c2920b96761c1/.pre-commit-config.yaml#L86-L102

If we want to enforce lists for instance, I have the feeling it would be easier to implement (and also let the possibility of unit testing it if needed).

Any opinion?

AcquaDiGiorgio: I used bash for the commodity of grep, but I can change the script to use python with pixi instead, it will definetely be much easier/cleaner to check its type that way.

chrisburr (changes requested):
This is a fake review for me to be able to test something 😄

chrisburr (changes requested):
Another fake review for me to be able to test something 😄

chrisburr (changes requested):
Another fake review for me to be able to test something 😄

chrisburr (changes requested):
Yet another fake review for me to be able to test something 😄

AcquaDiGiorgio:
I don't know how much of an improvement this is, but we might not want to use importlib for this task.

To import an `__init__.py` file, the only way I was able to find is doing:
```python
spec = importlib.util.spec_from_file_location(module_name, file)
module = importlib.util.module_from_spec(spec)
spec.loader.exec_module(module)
```
Here, `module_name` would be for example, `diracx.db.sql.utils` for the file `diracx-db/src/diracx/db/sql/utils/__init__.py`. With this, importlib is able to resolve relative paths. 

While importing the module, if the `__all__` dunder is defined after the imports, they are resolved first and raise an Exception if the module to be imported is not found. If we want this for the gubbins extensions, we will need to use the gubbins pixi environment, because `gubbins.db.sql.utils` does not exist in the default one. 

However, as a counterpart of this method, importing the module each time is quite slow, and I don't see any issues on using regex.

Also, I found a wild `__init__.py` file in the tests for the db: `diracx-db/tests/pilot_agents/__init__.py`. Is this intentional?

AcquaDiGiorgio:
Note: I keep the PR as a draft to not interrupt whatever test you are doing, but is ready to review.

.github/workflows/check_init_files_precommit_hook.py
```diff
@@ -0,0 +1,60 @@
...
+        with open(file, "r") as f:
+            content = f.read()
+
+        # __all__ dunder exists
+        if not re.search("__all__ =", content):
+            files_without_all.append(file)
+
+        # If exists, make sure its a list
+        elif not re.search("__all__ =[\S\s]*\[", content):
+            files_incorrect_format_all.append(file)
```
aldbr: The approach sounds a bit fragile: does it work if you have `__all__` on multiple lines for instance? Or tricky case (I have no idea why one would do that but it's allowed) like `__all__ +=` or `__all__: list[str]`.
What about using `ast` to parse the file? https://docs.python.org/3/library/ast.html

aldbr (changes requested):

.github/workflows/check_init_files_precommit_hook.py
```diff
@@ -0,0 +1,60 @@
+"""Pre-commit hook for checking __all__ dunder in __init__.py files."""
+
+from __future__ import annotations
+
+import re
+from glob import glob
+
+
+def main():
```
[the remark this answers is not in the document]
AcquaDiGiorgio: Yes, by not setting `pass_filenames: false`, but still, this could have been a simple sh script. I will change this because it makes more sense, but I would love to know why pixi is the better option.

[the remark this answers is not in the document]
aldbr: `pixi` can be seen as a unnecessary for this specific script. The main argument for keeping it is consistency with the rest of our tooling, so contributors don't have to remember that this hook is special and runs differently.

What's your opinion? How would you handle that?

aldbr:
> Also, I found a wild __init__.py file in the tests for the db: diracx-db/tests/pilot_agents/__init__.py. Is this intentional?

No, I don't think so. Can you remove it please?

aldbr:
Here is an example of what we want I think:
- general rule: consumers should import from submodules (e.g. `diracx.logic.auth import do_device_flow`, `diracx.core.model import TokenPayload`). We should only add modules to `__all__` if used elsewhere (we want to limit the public API).
- `diracx.(cli/api/client/core/routers/logic/db)`: `__all__ = []` because we don't want to import modules like `diracx.logic import device_flow`.
- `diracx.db.(sql/os)`: a bit special but here we keep as-is, consumers "don't get direct access to the modules under `sql/os`.

I didn't check all the sub directories and modules, it's just to give you a sense of what we should probably aim for.
Any opinion?

extensions/gubbins/gubbins-routers/src/gubbins/routers/well_known.py
```diff
@@ -6,11 +6,15 @@
 * uses the Gubbins dependencies
 """
 
-from diracx.routers.auth.well_known import router as diracx_wellknown_router
+from __future__ import annotations
+
+__all__ = ["get_installation_metadata"]
```
aldbr: Can you explain why you add the `__all__` directly in the `well_known` module here?
For `routers`, especially in the extension, I think we should not have anything (who would reuse the `routers` elsewhere?). What do you think?

aldbr (changes requested):

extensions/gubbins/gubbins-routers/src/gubbins/routers/well_known.py
AcquaDiGiorgio: `get_installation_metadata` is used outside `well_known.py`, so it needs to appear here, but I agree that `router` might be unnecesary on other files such as at `gubbins/routers/auth/__init__.py`.

extensions/gubbins/gubbins-routers/src/gubbins/routers/well_known.py
```diff
@@ -6,11 +6,15 @@
...
+
+from diracx.routers.auth import router as diracx_wellknown_router
```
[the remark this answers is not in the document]
AcquaDiGiorgio: Yeah, my bad. However, does this mean that `router` should appear at the `__all__` from `diracx/routers/auth/well_known.py`? Because then this doesn't follow the "convention".

.github/workflows/integration.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Clone DIRAC
         run: |
           pip install typer pyyaml gitpython packaging
-          git clone https://github.com/DIRACGrid/DIRAC.git -b "${{ matrix.dirac-branch }}" /tmp/DIRACRepo
+          git clone https://github.com/AcquaDiGiorgio/DIRAC.git -b "diracx-issue-426-__all__-in-__init__" /tmp/DIRACRepo
```
aldbr: Can you revert the changes here please?

.github/workflows/main.yml
```diff
@@ -168,7 +168,7 @@ jobs:
           docker image ls -a
       - name: Start demo
         run: |
-          git clone https://github.com/DIRACGrid/diracx-charts.git ../diracx-charts
+          git clone -b "diracx-issue-426-__all__-in-__init__" https://github.com/AcquaDiGiorgio/diracx-charts.git ../diracx-charts
```
aldbr: Same here

diracx-routers/src/diracx/routers/access_policies.py
```diff
@@ -17,6 +17,10 @@
 
 """
 
+from __future__ import annotations
+
+__all__ = ["BaseAccessPolicy", "check_permissions", "open_access"]
```
aldbr, line 22: This won't be checked by your script I think. Is that expected?

diracx-logic/tests/rss/__init__.py
aldbr, line 1: I don't think it's necessary to add `__all__` to the `tests` directories. What do you think?

diracx-routers/src/diracx/routers/access_policies.py
```diff
@@ -17,6 +17,10 @@
 
 """
 
+from __future__ import annotations
```
aldbr, line 20: Since you now have added that line, this may be removed in `pyproject.toml`, to investigate:

```
[tool.ruff.lint.extend-per-file-ignores]
"diracx-routers/src/diracx/routers/access_policies.py" = ["I002"]
```

.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ default_language_version:
   python: python3
 
 ci:
-  skip: [generate-pixi-docs, settings-doc-check]
+  skip: [generate-pixi-docs, settings-doc-check, check-init-files]
```
aldbr: I don't think we want to ignore that check in the CI.
What if you run with `language: python` and `entry: python ...` instead? You can mention that during Ddev to see what other devs think about that

aldbr (changes requested):
Looks almost ok. Just a few additional general comments:
- we could potentially add these ruff rules (complementary to your work):
  - https://docs.astral.sh/ruff/rules/invalid-all-object/
  - https://docs.astral.sh/ruff/rules/unsorted-dunder-all/

aldbr:
See https://github.com/DIRACGrid/diracx/pull/912/changes to potentially run your script with `pixi` in the CI (?)

diracx-logic/tests/rss/__init__.py
AcquaDiGiorgio: I don't think so either, I will add it to the exclude section

diracx-routers/src/diracx/routers/access_policies.py
AcquaDiGiorgio: Yeah, it only checks init files. 
I will look into this, because it should also look for the dunder in the files at the root of each project, now that you mention it.

diracx-routers/src/diracx/routers/access_policies.py
AcquaDiGiorgio: Yeah, it can be removed, since the pull request has been merged already.

AcquaDiGiorgio:
> Looks almost ok. Just a few additional general comments:
> 
> * we could potentially add these ruff rules (complementary to your work):
>       
>    * https://docs.astral.sh/ruff/rules/invalid-all-object/
>    * https://docs.astral.sh/ruff/rules/unsorted-dunder-all/

[invalid-all-object (PLE0604)](https://docs.astral.sh/ruff/rules/invalid-all-object/#invalid-all-object-ple0604) is already in use. In the `project.toml` we check for every PLE rule.
[unsorted-dunder-all (RUF022)](https://docs.astral.sh/ruff/rules/unsorted-dunder-all/#unsorted-dunder-all-ruf022) could be a nice-to-have. I can add it.

.pre-commit-config.yaml
aldbr: What about that? Have you tried to see if you could run it with `pixi`? Or else to run it with `python` to avoid skipping the CI?

AcquaDiGiorgio:
In principle, this should be it.
I don't know if this what you wanted in relation to pixi and pre-commit, but it now checks the `__all__` during CI execution too. I can make that only executes my hook, but I think including the `settings-doc-check` hook is a good idea.

aldbr (approved):

PR #776, opened 2026-02-09
fix: include LICENSE file in sub-package sdists and wheels

chrisburr:
## Summary
- Each sub-package (`diracx-api/`, `diracx-core/`, etc.) was missing the LICENSE file in published sdists and wheels because hatchling only auto-discovers license files in the package's own directory, not the monorepo root
- Added `[tool.hatch.build.targets.sdist.force-include]` to all 8 sub-package `pyproject.toml` files to pull the root `LICENSE` into each sdist
- The wheel (built from the sdist) then auto-discovers the LICENSE and places it in `.dist-info/licenses/`

### DIFF
--- a/diracx-api/pyproject.toml
+++ b/diracx-api/pyproject.toml
@@ -35,6 +35,9 @@ source = "vcs"
 [tool.hatch.version.raw-options]
 root = ".."
 
+[tool.hatch.build.targets.sdist.force-include]
+"../LICENSE" = "LICENSE"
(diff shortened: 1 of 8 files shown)


aldbr (commented):
Don't you need the same change for the `wheel` for completeness?

https://hatch.pypa.io/1.9/config/build/#forced-inclusion

aldbr (approved):

PR #777, opened 2026-02-09
feat(diracx-api): add set_job_status

Loxeris:
Adds the `set_job_status` function to set the status, minor status and application status of a job.

Needed for issue https://github.com/DIRACGrid/dirac-cwl/issues/83

### DIFF
--- a/diracx-api/src/diracx/api/jobs.py
+++ b/diracx-api/src/diracx/api/jobs.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-__all__ = ("create_sandbox", "download_sandbox")
+__all__ = ("create_sandbox", "download_sandbox", "set_job_status")
 
 import hashlib
 import logging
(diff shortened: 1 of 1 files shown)


diracx-api/src/diracx/api/jobs.py
```diff
@@ -123,3 +129,60 @@ async def download_sandbox(pfn: str, destination: Path, *, client: AsyncDiracCli
         with tarfile_open(fh) as tf:
             tf.extractall(path=destination, filter="data")
         logger.debug("Extracted %s to %s", pfn, destination)
+
+
+@with_client
+async def set_job_status(
+    job_id: str,
```
aldbr: I think `job_id` is expected to be an `int`
```suggestion
    job_id: int,
```

chaen:
why not just calling the client directly ? There does not seem to be any added logic here ?

Issue #642, opened 2025-08-26
date_format is slow


chaen:
Taken from the jobDB:

```sql
MySQL Dirac@dbod-lbwms.cern.ch:JobDB> select count(*) FROM `Jobs` WHERE `Jobs`.`Status` = 'Done' AND `Jobs`.`Site` = 'LCG.CERN.cern' AND `Jobs`.`LastUpdateTime` > '2025-08-25 15:35:31';
+----------+
| count(*) |
+----------+
| 56038    |
+----------+

1 row in set
Time: 0.649s
MySQL Dirac@dbod-lbwms.cern.ch:JobDB> select count(*) FROM `Jobs` WHERE `Jobs`.`Status` = 'Done' AND `Jobs`.`Site` = 'LCG.CERN.cern' AND date_format(`Jobs`.`LastUpdateTime`, '%Y-%m-%d %H:%i:%S') > '2025-08-25 15:35:31';
+----------+
| count(*) |
+----------+
| 56042    |
+----------+

1 row in set
Time: 12.806s
```

aldbr:
> What's the purpose?

Know if the job is newer than a given date

- Requires SQL proficiency
- Needs to be tested on a production jobDB

Will be done by someone from LHCb ideally because it's easier

PR #809, opened 2026-02-26
perf: use sargable (Search ARGument ABLE) range predicates for datetime search filters

ryuwd:
- `apply_search_filters` previously used `date_trunc()` to wrap datetime columns in `date_format()` (MySQL) / `strftime()` (SQLite), which prevents the database from using indexes on those columns
- Replaced with range-based comparisons on the raw column (e.g., `col >= start AND col < end` instead of `date_format(col, '%Y-%m-%d') = '2025-08-25'`)
- All precision levels (YEAR through SECOND) and all comparison operators (eq, neq, gt, lt, in, not in) are handled with index-friendly range predicates

```sql
-- Direct comparison (uses index): 0.649s
SELECT count(*) FROM Jobs
WHERE Status = 'Done' AND Site = 'LCG.CERN.cern'
  AND LastUpdateTime > '2025-08-25 15:35:31';

-- date_format() wrapper (full table scan): 12.806s
SELECT count(*) FROM Jobs
WHERE Status = 'Done' AND Site = 'LCG.CERN.cern'
  AND date_format(LastUpdateTime, '%Y-%m-%d %H:%i:%S') > '2025-08-25 15:35:31';
```

Closes https://github.com/DIRACGrid/diracx/issues/642

### DIFF
--- a/diracx-db/src/diracx/db/sql/utils/base.py
+++ b/diracx-db/src/diracx/db/sql/utils/base.py
@@ -421,11 +421,11 @@ def _build_datetime_range_expr(
     """
     if operator == ScalarSearchOperator.EQUAL:
         return and_(column >= start, column < end)
-    elif operator == ScalarSearchOperator.NOT_EQUAL:
+    if operator == ScalarSearchOperator.NOT_EQUAL:
         return or_(column < start, column >= end)
-    elif operator == ScalarSearchOperator.GREATER_THAN:
+    if operator == ScalarSearchOperator.GREATER_THAN:
         return column >= end
-    elif operator == ScalarSearchOperator.LESS_THAN:
+    if operator == ScalarSearchOperator.LESS_THAN:
         return column < start
     raise InvalidQueryError(
         f"Operator '{operator}' is not supported for partial datetime values"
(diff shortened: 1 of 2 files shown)


diracx-db/src/diracx/db/sql/utils/base.py
```diff
@@ -359,6 +369,84 @@ def _get_columns(table, parameters):
...
+def _build_datetime_range_multi_expr(
+    column,
+    operator: VectorSearchOperator,
+    bounds: list[tuple[datetime, datetime]],
+):
+    """Build a sargable range expression for multiple datetime periods (IN/NOT IN)."""
+    if operator == VectorSearchOperator.IN:
+        return or_(*[and_(column >= s, column < e) for s, e in bounds])
+    elif operator == VectorSearchOperator.NOT_IN:
+        return and_(*[or_(column < s, column >= e) for s, e in bounds])
```
aldbr:
```suggestion
    if operator == VectorSearchOperator.IN:
        return or_(*[and_(column >= s, column < e) for s, e in bounds])
    if operator == VectorSearchOperator.NOT_IN:
        return and_(*[or_(column < s, column >= e) for s, e in bounds])
```

PR #811, opened 2026-02-27
feat(core): add generic StateMachine base class for typed state transitions

ryuwd:
Reusable declarative state machine utility for diracx. Each system subclasses StateMachine with StrEnum states and a transitions dict. Validation runs at class creation via __init_subclass__.


aldbr:
Is there any github issue linked to your PR?
Can you briefly explain why you need that, and why now please?

PR #826, opened 2026-03-10
chore: added API validation for numeric values

### LINKED ISSUE #506: Input validation is required to avoid 500 errors
In every endpoint where we handle numbers, we must verify that the input is less than a certain (fixed) maximum.

Sometimes, the db raises 500 errors for multiple reasons:

- In the job query for example `pymysql.err.ProgrammingError`, setting $\text{page number }>100$
- In a PR I'm doing for Pilots where I insert an int in the db, I have a `pymysql.err.DataError) (1264, "Out of range value for column ...")`

I did not check every endpoint, but we should have for every type that we input in an endpoint and that will be used in the DB, something to check if it's valid or not.

Stellatsuu:
Closes: https://github.com/DIRACGrid/diracx/issues/506

Changes:
- added min-max validation in `per_page` query value `/jobs/search`
- added min validation in `page` query value of `/jobs/search`
- updated tests

Comments:
I can't reproduce this error from the issue, I think it has been fixed somewhere else:
> In the job query for example pymysql.err.ProgrammingError, setting page number  > 100 

I don't really see other cases here for numeric values. Most numeric values are `job_ids` that are stored in DB and can't be validated from the FastAPI schemas, same for the `page` max value that can't be validated since it's based on DB, anyways, a `page` value out of bound will return a 416 error = no jobs found at this `page` value (might be why I can't reproduce the error above).

As for strings values, I did not find any relevant cases, most relevant ones does already have a validation (such as `max_length` and `regex`)


aldbr (commented):
Thanks @Stellatsuu!
About the client generation, are you using the same version locally than the one in the CI? (normally yes, but worth checking)

chrisburr (approved):

PR #842, opened 2026-03-20
feat: add diracx-tasks

chrisburr:


diracx-routers/src/diracx/routers/dependencies.py
```diff
@@ -1,72 +1,7 @@
...
-
-AuthSettings = Annotated[_AuthSettings, Depends(_AuthSettings.create)]
-DevelopmentSettings = Annotated[
-    _DevelopmentSettings, Depends(_DevelopmentSettings.create)
-]
-SandboxStoreSettings = Annotated[
-    _SandboxStoreSettings, Depends(_SandboxStoreSettings.create)
-]
+# Re-export everything from the canonical location
+from diracx.tasks.plumbing.depends import *  # noqa: F401, F403
```
aldbr: Do we want `diracx-routers` to depend on `diracx-tasks`?
I would have naively imagined:
`diracx-logic`  
|
  -> `diracx-tasks`
|
  -> `diracx-routers`

PR #857, opened 2026-03-25
feat: add RSS data models (SQLAlchemy + Pydantic)

### LINKED ISSUE #837: RSS data models (SQLAlchemy + Pydantic)
### Description

Create the data layer for RSS in diracx:
1. **SQLAlchemy 2.0 models** in `diracx-db` that map the existing `ResourceStatusDB` tables — no schema changes, DB is shared with legacy DIRAC.
2. **Pydantic models** in `diracx-core` for the consumer-facing view — a simplified binary allowed/banned model.

### SQLAlchemy models

> **Note**: The `ResourceStatusCache` table from the legacy schema is not needed — caching is handled by `CacheableSource` (issue C).

**Technical details:**
- Create `diracx-db/src/diracx/db/sql/rss/` package with `schema.py` and `db.py`
- Follow diracx SQLAlchemy 2.0 patterns (`DeclarativeBase`, `Mapped[T]`, `mapped_column()`) — see `diracx-db/src/diracx/db/sql/auth/schema.py` and `job/schema.py` for reference
- Create `ResourceStatusDB(BaseSQLDB)` with read methods for status queries
- Register via entry point in `pyproject.toml` under `diracx.dbs.sql`
- Must **not** create or alter tables — schema is owned by legacy DIRAC

### Pydantic models

Implement the simplified binary consumer model proposed in the [#790 discussion](https://github.com/DIRACGrid/diracx/issues/790#issuecomment-4012486144):

```python
from pydantic import BaseModel
from typing import Literal

class AllowedStatus(BaseModel):
    allowed: Literal[True]
    warnings: list[str] = []

    def __bool__(self) -> bool:
        return True

class BannedStatus(BaseModel):
    allowed: Literal[False]
    reason: str = "Unknown"

    def __bool__(self) -> bool:
        return False

ResourceStatus = AllowedStatus | BannedStatus  # Discriminated union on `allowed`

class StorageElementStatus(BaseModel):
    read: ResourceStatus
    write: ResourceStatus
    check: ResourceStatus
    remove: ResourceStatus

class ComputeElementStatus(BaseModel):
    all: ResourceStatus

class SiteStatus(BaseModel):
    all: ResourceStatus
```

**Mapping from DB status to binary model:**
- **Allowed**: `Active`, `Degraded`
- **Banned**: `Banned`, `Probing`, `Error`, `Unknown`

This matches the existing `getUsableSites` logic in DIRAC (`SiteStatus.py`).

### Reference

- Legacy schema: [`DIRAC/ResourceStatusSystem/DB/ResourceStatusDB.py`](https://github.com/DIRACGrid/DIRAC/blob/main/src/DIRAC/ResourceStatusSystem/DB/ResourceStatusDB.py) (lines 86–260)
- Chris's model proposal: [#790 comment](https://github.com/DIRACGrid/diracx/issues/790#issuecomment-4012486144)
- Meeting agreement: all resource types (Storage, Compute, Site) implemented, binary allowed/banned for consumers

### Acceptance criteria

- [ ] SQLAlchemy 2.0 models for the tables, matching the existing schema exactly
- [ ] `ResourceStatusDB(BaseSQLDB)` class with read methods for `SiteStatus` and `ResourceStatus`
- [ ] Entry point registered in `pyproject.toml`
- [ ] Pydantic models for `AllowedStatus`, `BannedStatus`, `StorageElementStatus`, `ComputeElementStatus`, `SiteStatus`
- [ ] `ResourceType` enum
- [ ] Unit tests for DB reads and model serialization/deserialization
- [ ] No DB migrations or schema changes

HeloiseJoffe:
Closes #837 

Changes : 
- Add RSS SQLAlchemy 2.0 models for the tables and ResourceStatusDB class
- Add RSS entry point
- Add pydantic models and ResourceType enum
- Add Unit tests


diracx-db/src/diracx/db/sql/rss/schema.py
```diff
@@ -0,0 +1,115 @@
...
+class ResourceHistory(ElementStatusBaseWithID, RSSBase):
+    __tablename__ = "ResourceHistory"
+
+
+class NodeLog(ElementStatusBaseWithID, RSSBase):
+    __tablename__ = "NodeLog"
+
+
+class NodeHistory(ElementStatusBaseWithID, RSSBase):
+    __tablename__ = "NodeHistory"
```
fstagni: These 2 have been removed, no need to add them here
```suggestion
```

diracx-core/src/diracx/core/models/rss.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+ResourceStatus = Annotated[
+    Union[AllowedStatus, BannedStatus],
+    Field(discriminator="allowed"),
+]
+
+
+class ResourceType(StrEnum):
+    Site = "Site"
```
fstagni: "Site" is not a "Resource" in RSS terms
```suggestion
```

aldbr (changes requested):

diracx-db/src/diracx/db/sql/rss/schema.py
```diff
@@ -0,0 +1,103 @@
...
+    elementtype: Mapped[str32] = mapped_column("ElementType", server_default="")
+    lastchecktime: Mapped[datetime] = mapped_column(
+        "LastCheckTime", DateTime, server_default="1000-01-01 00:00:00"
+    )
+    tokenowner: Mapped[str] = mapped_column(
+        "TokenOwner", String(16), server_default="rs_svc"
+    )
+
+
+class ElementStatusBaseWithID(ElementStatusBase):
```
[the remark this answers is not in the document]
fstagni: Maybe we could have only one type, but that would mean changing the table definitions in DIRAC too, which as we know implies manual operation. I would keep as it is for now.

diracx-logic/src/diracx/logic/rss/query.py
aldbr, line 1: What about calling that file `query.py`?

diracx-core/src/diracx/core/models/rss.py
```diff
@@ -0,0 +1,73 @@
...
+    if db_status in BANNED:
+        return BannedStatus(
+            allowed=False,
+            reason=reason or db_status,
+        )
+
+    return BannedStatus(
+        allowed=False,
+        reason=f"Unknown status: {db_status}",
+    )
```
aldbr: What about moving `map_status` to `diracx-logic`?

diracx-logic/tests/rss/test_rss.py
aldbr, line 1: And therefore you can potentially move that test to `diracx-logic` too

diracx-db/src/diracx/db/sql/rss/schema.py
```diff
@@ -0,0 +1,101 @@
...
+    type_annotation_map = {
+        str32: String(32),
+        str64: String(64),
+        str128: String(128),
+        str512: String(512),
+    }
+
+
+class ElementStatusBase:
+    __table_args__ = {"mysql_engine": "InnoDB", "mysql_charset": "utf8mb4"}
```
aldbr: Do you really need that?
```suggestion
```

diracx-db/src/diracx/db/sql/rss/schema.py
```diff
@@ -0,0 +1,101 @@
...
+
+    name: Mapped[str64] = mapped_column("Name", primary_key=True)
+    statustype: Mapped[str128] = mapped_column(
+        "StatusType", server_default="all", primary_key=True
+    )
+    vo: Mapped[str64] = mapped_column("VO", primary_key=True, server_default="all")
+    status: Mapped[str] = mapped_column("Status", String(8), server_default="")
+    reason: Mapped[str512] = mapped_column("Reason", server_default="Unspecified")
+    dateeffective: Mapped[datetime] = mapped_column("DateEffective", DateTime)
```
aldbr: Don't you need to pass `SmarterDateTime` here? Same for the other dates.
```suggestion
    date_effective: Mapped[datetime] = mapped_column("DateEffective", SmarterDateTime())
```

diracx-db/src/diracx/db/sql/rss/schema.py
```diff
@@ -0,0 +1,101 @@
...
+    Differences:
+    - there's an autoincrement ID column which is also the primary key
+    - the name and statusType components are not part of the primary key
+    """
+
+    id: Mapped[int] = mapped_column(
+        "ID", BigInteger, autoincrement=True, primary_key=True
+    )
+    name: Mapped[str64] = mapped_column("Name")
+    statustype: Mapped[str128] = mapped_column("StatusType", server_default="all")
```
aldbr: Any reason for not using snake case as in the other table definitions? Same for all the attribute names
```suggestion
    status_type: Mapped[str128] = mapped_column("StatusType", server_default="all")
```

diracx-core/src/diracx/core/models/rss.py
```diff
@@ -0,0 +1,73 @@
...
+    all: ResourceStatus
+
+
+ALLOWED = {"Active", "Degraded"}
+BANNED = {"Banned", "Probing", "Error", "Unknown"}
+
+
+def map_status(db_status: str, reason: str | None = None) -> ResourceStatus:
+    if db_status in ALLOWED:
+        return AllowedStatus(allowed=True)
```
aldbr: What about the `warnings`?

diracx-db/tests/rss/test_rss_db.py
```diff
@@ -0,0 +1,135 @@
...
+                DateEffective=_NOW,
+                TokenExpiration=_FAR,
+                LastCheckTime=_NOW,
+                ElementType="Site",
+                TokenOwner="test",
+            )
+        )
+
+    # Test with the test Site (should be found)
+    async with rss_db as rss_db:
```
aldbr: Minor: to avoid name shadowing
```suggestion
    async with rss_db as db:
```

diracx-db/src/diracx/db/sql/rss/db.py
```diff
@@ -0,0 +1,52 @@
...
+    SiteStatus,
+)
+
+
+class ResourceStatusDB(BaseSQLDB):
+    """Class that defines the interactions with the tables of the ResourceStatusDB."""
+
+    metadata = RSSBase.metadata
+
+    async def get_site_status(self, name: str, vo: str = "all") -> tuple[str, str]:
```
aldbr, line 21: @fstagni out of curiosity, shouldn't we (not now but later) transition from this magic `all` to NULL in the DB?
If not VO is set, then we assume it's for all of them.
Unless I miss something?

diracx-db/src/diracx/db/sql/rss/db.py
```diff
@@ -0,0 +1,52 @@
...
+        row = result.one_or_none()
+        if not row:
+            raise ResourceNotFoundError(name)
+
+        return row.Status, row.Reason
+
+    async def get_resource_status(
+        self,
+        name: str,
+        statustypes: list[str] = ["all"],
```
aldbr: Beware of https://docs.astral.sh/ruff/rules/mutable-argument-default/
It's fine here because we don't mutate `status_types` but we should be careful.

@fstagni same here, should we transition from `all` for VO and StatusType to NULL at some point?

aldbr (changes requested):

diracx-core/src/diracx/core/models/rss.py
fstagni: The should be what is in `reason`, at least for the case when `db_status` is "Degraded".

diracx-db/src/diracx/db/sql/rss/db.py
fstagni: I would agree with you

diracx-db/src/diracx/db/sql/rss/db.py
fstagni: I would agree to this one too.

diracx-core/src/diracx/core/exceptions.py
```diff
@@ -70,6 +70,12 @@ def __init__(self, pfn: str, se_name: str, detail: str | None = None):
         )
 
 
+class ResourceNotFoundError(DiracError):
```
fstagni, line 73: I am really wondering if we need this one at all.

diracx-db/src/diracx/db/sql/rss/schema.py
HeloiseJoffe: No, I don't think so. It comes from the DIRAC version.

diracx-db/src/diracx/db/sql/rss/schema.py
HeloiseJoffe: Same, it's a copy from DIRAC, I will adapt it

diracx-core/src/diracx/core/exceptions.py
HeloiseJoffe: I used to use a `raise ValueError` to handle missing sites and resources, but Alexandre advised me to use a `ResourceNotFoundError` in `diracx/core/exceptions.py` instead. 
Do you think we should handle it differently ?

aldbr (approved):

PR #883, opened 2026-04-08
feat: add task to clean sandbox store

chrisburr:
Rebased the work of @natthan-pigoux based on how tasks evolved as a result of his original PR (https://github.com/chrisburr/diracx/pull/1).

### DIFF
--- a/diracx-tasks/src/diracx/tasks/jobs/clean_sandbox_store.py
+++ b/diracx-tasks/src/diracx/tasks/jobs/clean_sandbox_store.py
@@ -1,0 +1,31 @@
+"""Task definition to clean Sandbox Store."""
+
+from __future__ import annotations
+
+import dataclasses
+from typing import Annotated
+
+from diracx.core.settings import SandboxStoreSettings
+from diracx.db.sql import SandboxMetadataDB
+from diracx.logic.jobs.sandboxes import clean_sandboxes
+from diracx.tasks.plumbing.base_task import PeriodicBaseTask
+from diracx.tasks.plumbing.enums import Priority, Size
+from diracx.tasks.plumbing.retry_policies import ExponentialBackoff
+from diracx.tasks.plumbing.schedules import CronSchedule
+
+from .depends import NoTransaction
+
+
+@dataclasses.dataclass
+class CleanSandboxStoreTask(PeriodicBaseTask):
+    priority = Priority.BACKGROUND
+    size = Size.MEDIUM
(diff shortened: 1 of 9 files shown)


aldbr:
Shall we add a unit test?

chrisburr:
`diracx.logic.jobs.sandboxes.clean_sandboxes` is already tested so here we have a question how we want to handle testing tasks.

I'm not sure what the answer should be for cases like this where tasks is just minimal boiler plate.

diracx-tasks/src/diracx/tasks/jobs/clean_sandbox_store.py
```diff
@@ -0,0 +1,31 @@
...
+
+from diracx.core.settings import SandboxStoreSettings
+from diracx.db.sql import SandboxMetadataDB
+from diracx.logic.jobs.sandboxes import clean_sandboxes
+from diracx.tasks.plumbing.base_task import PeriodicBaseTask
+from diracx.tasks.plumbing.enums import Priority, Size
+from diracx.tasks.plumbing.retry_policies import ExponentialBackoff
+from diracx.tasks.plumbing.schedules import CronSchedule
+
+from .depends import NoTransaction
```
chrisburr: This looks buggy...

chaen (changes requested):
buggy import

chrisburr:
This branch had a broken import in `CleanSandboxStoreTask` as it imported `NoTransaction` from a `.depends` module that didn't exist. This was missed in CI because:

  1. The jobs/ directory was missing `__init__.py`, so ruff/mypy didn't analyse the file
  2. No test imported the module or loaded the entry point
  3. Entry points are resolved lazily, so the `ImportError` only surfaces at runtime

I've added several commits on top to fix this and prevent it happening again:

  - Enable ruff INP001 to catch missing `__init__.py` files (added `__init__.py` to all test subdirectories and `diracx-tasks/src/diracx/tasks/jobs/`)
  - Create `diracx.tasks.depends` as the public re-export module for dependency markers (`NoTransaction`, `CallbackSpawner`), so task authors don't need to reach into `plumbing` internals
  - Fix the broken import in `CleanSandboxStoreTask` and populate `jobs/__init__.py`
  - Add a parametrized test that calls `.load()` on every registered `diracx.*` entry point
  - Update docs to reference `diracx.tasks.depends` instead of `diracx.tasks.plumbing.depends`

chaen (approved):

PR #885, opened 2026-04-13
fix: drop support for EdDSA keys

### LINKED ISSUE #806: Remove `EdDSA` algorithm from the possible JWK values
Following https://github.com/DIRACGrid/diracx/pull/786, once released, communities will have to rotate their keys.
Once the process is done, one will have to remove `EdDSA` from the algorithms:

https://github.com/DIRACGrid/diracx/blob/84c7651eb644947009f10ce3fa3ce4a788a9b8a7/diracx-core/src/diracx/core/settings.py#L207-L208

aldbr:
Closes https://github.com/DIRACGrid/diracx/issues/806
Dropping support for EdDSA algorithm name for good.

### DIFF
--- a/diracx-core/src/diracx/core/settings.py
+++ b/diracx-core/src/diracx/core/settings.py
@@ -227,8 +227,7 @@ class AuthSettings(ServiceSettingsBase):
     generation and verification.
     """
 
-    # TODO: EdDSA should be removed later due to "SecurityWarning: EdDSA is deprecated via RFC 9864"
-    token_allowed_algorithms: list[str] = ["RS256", "EdDSA", "Ed25519"]  # noqa: S105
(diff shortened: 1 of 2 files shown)


chrisburr:
My understanding is that this has been fixed in LHCb?

Are there any other installations that need to be updated before this can be merged?

aldbr:
> My understanding is that this has been fixed in LHCb?

Yes

> Are there any other installations that need to be updated before this can be merged?

@Stellatsuu is there anything to do in your installation?

natthan-pigoux:
> @Stellatsuu is there anything to do in your installation?

@aldbr , all good from our side, thanks.

chrisburr (approved):

PR #887, opened 2026-04-14
feat: add Docker image for tasks

chrisburr:
The container-tasks pixi environment was already defined but no Docker image was being built or published for it. Add build steps to both the deployment and CI workflows so tasks workers/schedulers get their own dedicated image.


aldbr (changes requested):
I guess you first need https://github.com/DIRACGrid/diracx-charts/pull/261 no?
Also don't you need to edit the `gubbins-charts values.yaml` file to include the `gubbins/tasks` image?

chaen:
no, we first need the images before the charts

fstagni (approved):

PR #902, opened 2026-04-22
fix: stop using UPGRADE_REQUIRED for client version check

### LINKED ISSUE #542: Stop using UPGRADE_REQUIRED
It's intended for HTTP protocol changes rather than client version upgrades.

HeloiseJoffe:
Closes #542 

Change:
Replace UPGRADE_REQUIRED (426) with PRECONDITION_FAILED (412)

### DIFF
--- a/diracx-routers/src/diracx/routers/factory.py
+++ b/diracx-routers/src/diracx/routers/factory.py
@@ -508,10 +508,9 @@ class ClientMinVersionCheckMiddleware(BaseHTTPMiddleware):
                 # This is not managed here.
 
                 raise HTTPException(
-                    status_code=HTTPStatus.UPGRADE_REQUIRED,
-                    detail=f"Client version ({client_version})"
-                    f"not recent enough (>= {self.min_client_version})."
-                    "Upgrade.",
+                    status_code=HTTPStatus.PRECONDITION_FAILED,
+                    detail=f"Client version {client_version} is not compatible with the server. "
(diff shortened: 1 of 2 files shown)


aldbr:
In the issue, we agreed on using `BAD_REQUEST`, can you explain why you decided to use `PRECONDITION_FAILED`?

HeloiseJoffe:
I saw `BAD_REQUEST` as an example rather than a strict choice. I used `PRECONDITION_FAILED` because the request itself is valid, but it does not satisfy the requirement enforced by the server, which felt closer to a precondition than a malformed request.

That said, I’m fine switching to `BAD_REQUEST` if we prefer keeping it simpler and more general.

aldbr:
Thanks for the explanation, it's true that `BAD_REQUEST` does not appear as the final choice from my comment in the issue indeed. The reasoning makes sense, but I would still lean toward `BAD_REQUEST`:
- `412 PRECONDITION_FAILED` does not seem to fully match with our case: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/412

  > The HTTP 412 Precondition Failed client error response status code indicates that access to the target resource was denied. This happens with conditional requests on methods other than GET or HEAD when the condition defined by the If-Unmodified-Since or If-Match headers is not fulfilled. In that case, the request (usually an upload or a modification of a resource) cannot be made and this error response is sent back.

- `400` seems more generic and details would come from the returned JSON.

aldbr (approved):

PR #906, opened 2026-04-29
fix: handle device flow errors

### LINKED ISSUE #559: Auth endpoint do not catch errors
Endpoint `do_device_flow` (`GET /device`) do not verify if an error was caught. It's is not important, but fixing will avoid 5XX errors.

By using a bad `user_code` will lead to an error (from the db, and later also from the parse scope).

Partially linked to #506 as we don't verify inputs and outputs.

Image

By the way, do we need to keep `diracx.routers.auth.util`? (`has_properties` function)

HeloiseJoffe:
Closes #559 

Changes:
- Replace the `NotImplementedError` with `IAMServerError` in the get_server_metadata function
- Add error handling in the `do_device_flow` routers endpoint for `NoResultFound`, `ValueError` and `IAMServerError`

Question: 
The issue mentioned: "do we need to keep `diracx.routers.auth.util`? (`has_properties` function)"
After looking at the codebase, I think `has_properties` is not needed since access policies (`WMSAccessPolicy`, `SandboxAccessPolicy` ...) already handle property checks.
However, I'm not sure if the removing of this function should be done here or in a separate PR?
What do you think about it ?


diracx-routers/src/diracx/routers/auth/device_flow.py
```diff
@@ -101,14 +102,33 @@ async def do_device_flow(
...
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=e.args[0],
+        ) from e
+    except IAMServerError as e:
+        logger.warning("IAM server error during device flow: %s", e)
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail=e.args[0],
+        ) from e
```
aldbr, line 131: I assume these cases are not unit tested, am I wrong?
Could you add some unit tests for them please?

aldbr:
> After looking at the codebase, I think has_properties is not needed since access policies (WMSAccessPolicy, SandboxAccessPolicy ...) already handle property checks.

I think you are right indeed, it's not used anymore anywhere, just referenced in the document and I presume it should be deleted. 

@chaen @chrisburr any opinion?

aldbr:
> > After looking at the codebase, I think has_properties is not needed since access policies (WMSAccessPolicy, SandboxAccessPolicy ...) already handle property checks.
> 
> I think you are right indeed, it's not used anymore anywhere, just referenced in the document and I presume it should be deleted.
> 
> @chaen @chrisburr any opinion?

After discussing with @chaen, we agreed that it could be deleted, and I think you can do it as part of this PR.
Is that ok for you @HeloiseJoffe?

aldbr (approved):

PR #917, opened 2026-05-19
refactor: migrate fields to snake_case with alias

### LINKED ISSUE #618: Need consistency in variable naming between DB fields and Python models
We defined several classes to type both the payload of the endpoint and its returned value.
https://github.com/DIRACGrid/diracx/blob/b913bf4b9f65b49c53f075e031fb249494fbd5ed/diracx-core/src/diracx/core/models.py#L128-L141
On the database side, the naming convention for fields is **PascalCase**.
On the Python side, the convention is **snake_case**.

## The issue

The user must refer to attributes using their **PascalCase** names (e.g., `JobType`). However, in Python, this is not standard practice — so we typically define fields like this:
```
job_type: str | None = Field(None, alias="JobType")
```

With this, `job_type` is the Python field name, and `JobType` is the alias used for external data (like JSON or the DB).
The problem is: when we try to use the Python name (`job_type`) to query or patch the DB, we mistakenly use `job_type` instead of `JobType`, which doesn't exist in the DB schema.

## Solution

- **Naive solution** (initially implemented):
    Don't follow Python naming conventions — use the `PascalCase` names directly in Python. That way, user input matches exactly the DB field name.
If we do that, we should add this choice in the ADR: https://github.com/DIRACGrid/diracx/issues/588 
    ➤ Downside: This breaks Python conventions, and becomes impossible when fields contain special characters (e.g., `Memory(MB)`).

- **Proper solution** (implemented here):
    For each field:
    - Use a proper Python **snake_case** name.
    - Define the actual DB field name as an alias via `Field(..., alias="PascalCaseName")`.

    However, **FastAPI automatically maps input data from alias to the Python name**.
    So if a user sends a request with `Memory(MB)`, FastAPI will map it to `memory_mb`.
    But our logic expects to operate on the original DB field names (i.e., aliases).

    ➤ To counter this, we need to re-map the internal Python names back to their alias when processing the payload. This ensures consistency with the DB field names.

## Remaining question

For consistency, should we update all existing classes to follow this approach (using `snake_case` names with aliases for DB fields), even if the fields don't contain special characters?

HeloiseJoffe:
Closes #618 

The FastAPI mapping is not blocking, since this can be handled correctly with the `by_alias=True` parameter.
I migrated the classes from `diracx/core/models/job.py` to snake_case with alias but not the other files from `diracx/core/models/` because they are already in snake_case.

I also started working on other `BaseModel` usages still defined in PascalCase, including : `diracx/cli/internal/legacy` and `diracx/core/config/schema.py`
However, after re-reading the issue, I am unsure whether these should also be migrated as part of this scope.
In the issue the discussion is about the python models and the DB fields.
Could you confirm @aldbr the intended scope of #618 ?
Should all remaining `BaseModel` classes be migrated to snake_case + aliases for consistency? Or is the goal limited to models directly tied to DB field mapping ?


aldbr (commented):
What about https://github.com/DIRACGrid/diracx/blob/06b9cee080c85dfb87bc7ebd1f1aeb8eb575448e/diracx-routers/src/diracx/routers/jobs/status.py#L111?

I guess extensions would have to modify their pydantic models if they already depend on them.
I suppose it's ok because we don't have many extensions for now and they are not very advanced, but we should likely check with the installation admins first.

aldbr:
> Could you confirm @aldbr the intended scope of https://github.com/DIRACGrid/diracx/issues/618 ?
> Should all remaining BaseModel classes be migrated to snake_case + aliases for consistency? Or is the goal limited to models directly tied to DB field mapping ?

If it's not too much, I would move all the `BaseModel` subclasses to snake case (alias only when necessary I would say, so only when the model is used to interact with the DB, or when it interacts with the end users?)

HeloiseJoffe:
It's done, I've migrated everything that interacts with the DB or end users. I hope I haven't missed anything. I also fixed the missing `by_alias` in the router.

aldbr (approved):

PR #922, opened 2026-05-29
feat: support global OS prefix

### LINKED ISSUE #8489: [Feature] Introduce the concept of a global prefix for OpenSearch indexes
The idea is to be able to specify a prefix which is always applied to the index created, and which is controlled by configuration. This is useful for example for installations having their OS cluster administered for them, and they are given a specific prefix they have to respect as security measure. 

There was the specifics of `Systems/Monitoring/Databases/MonitoringDB/IndexPrefix`  which could do that. That feature was removed in v9.0.

I propose to re-introduce it and generalize it: it could be defined for any OS DB, including `Systems/NoSQLDatabases/IndexPrefix`

The equivalent would be done in `diracx`, though I am not sure yet where to add that piece of configuration

@marianne013 @bertrandrigaud

chaen:
Needs https://github.com/DIRACGrid/diracx/pull/925

Closes https://github.com/DIRACGrid/DIRAC/issues/8489

### DIFF
--- a/diracx-core/src/diracx/core/settings.py
+++ b/diracx-core/src/diracx/core/settings.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 __all__ = [
     "AuthSettings",
     "DevelopmentSettings",
+    "FactorySettings",
     "LocalFileUrl",
     "SandboxStoreSettings",
     "ServiceSettingsBase",
@@ -389,6 +390,11 @@ class FactorySettings(ServiceSettingsBase):
         default="redis://localhost", validation_alias="DIRACX_TASKS_REDIS_URL"
     )
     """The url for the redis server to manage tasks"""
+
+    os_global_prefix: str = Field(
+        default="", validation_alias="DIRACX_FACTORY_OS_GLOBAL_PREFIX"
+    )
+    """Global prefix for OpenSearch database indices."""
 
     enabled_services: dict[str, bool] = Field(default_factory=dict)
     """The following environment variables dictates which routers are enabled."""
--- a/diracx-db/src/diracx/db/os/utils.py
+++ b/diracx-db/src/diracx/db/os/utils.py
@@ -80,13 +80,17 @@ class BaseOSDB(metaclass=ABCMeta):
     @abstractmethod
     def index_name(self, vo: str, doc_id: int) -> str: ...
 
-    def __init__(self, connection_kwargs: dict[str, Any]) -> None:
+    def __init__(
+        self, connection_kwargs: dict[str, Any], *, global_prefix: str = ""
+    ) -> None:
         self._client: AsyncOpenSearch | None = None
         self._connection_kwargs = connection_kwargs
         # We use a ContextVar to make sure that self._conn
         # is specific to each context, and avoid parallel
         # route executions to overlap
         self._conn: ContextVar[bool] = ContextVar("_conn", default=False)
+        if global_prefix:
+            self.index_prefix = f"{global_prefix}_{self.index_prefix}"
 
     @classmethod
     def available_implementations(cls, db_name: str) -> list[type[BaseOSDB]]:
--- a/diracx-db/src/diracx/db/os/utils.py
+++ b/diracx-db/src/diracx/db/os/utils.py
@@ -81,13 +81,17 @@ class BaseOSDB(metaclass=ABCMeta):
     @abstractmethod
     def index_name(self, vo: str, doc_id: int) -> str: ...
 
-    def __init__(self, connection_kwargs: dict[str, Any]) -> None:
+    def __init__(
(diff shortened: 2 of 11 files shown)


toto
aldbr, line 1: This file seems unexpected 🥲

diracx-testing/src/diracx/testing/mock_osdb.py
aldbr, line 1: How complex would it be to add a unit test against the mock OS DB?

diracx-tasks/src/diracx/tasks/plumbing/factory.py
```diff
@@ -305,9 +305,12 @@ async def setup_dependency_overrides(
                 )
 
         # --- OS databases ---
+        os_global_prefix = FactorySettings().os_global_prefix
```
aldbr, line 308: I guess you can create a variable to encapsulate a `FactorySettings` instance so that you can reuse it with `config_backend_url`.

diracx-core/src/diracx/core/settings.py
```diff
@@ -390,6 +391,11 @@ class FactorySettings(ServiceSettingsBase):
     )
     """The url for the redis server to manage tasks"""
 
+    os_global_prefix: str = Field(
```
aldbr, line 394: From what I understand, the prefix must be lower case.

aldbr (changes requested):

diracx-core/src/diracx/core/settings.py
chaen: Not sure I understand. Where is the upper case  ?

aldbr: Sorry it's indeed confusing even for me, I think I meant that there is nothing preventing one from adding a global prefix with upper case .

For instance here we use `lower` to control the input coming from outside: https://github.com/DIRACGrid/diracx/blob/712f32e9695610bba2f37de375bdf2f4a67195e6/diracx-db/src/diracx/db/os/job_parameters.py#L35

diracx-testing/src/diracx/testing/mock_osdb.py
chaen: The test can only be limited to making sure it is propagated to some extend but yes, I'll do so

diracx-core/src/diracx/core/settings.py
chaen: yes but `vo` is used in many context, while the `prefix` is something given by the opensearch admins. If it is with upper case, they have bigger issues :-)

diracx-db/src/diracx/db/os/utils.py
```diff
@@ -81,13 +81,17 @@ class BaseOSDB(metaclass=ABCMeta):
...
+        self, connection_kwargs: dict[str, Any], *, global_prefix: str = ""
+    ) -> None:
         self._client: AsyncOpenSearch | None = None
         self._connection_kwargs = connection_kwargs
         # We use a ContextVar to make sure that self._conn
         # is specific to each context, and avoid parallel
         # route executions to overlap
         self._conn: ContextVar[bool] = ContextVar("_conn", default=False)
+        if global_prefix:
+            self.index_prefix = f"{global_prefix}_{self.index_prefix}"
```
aldbr, line 94: Based on what's already in DIRAC: 

```suggestion
            self.index_prefix = f"{global_prefix}{self.index_prefix}"
```

See https://github.com/DIRACGrid/DIRAC/blob/aaa32ededf237de198a262fad76d292cee684290/src/DIRAC/Core/Utilities/ElasticSearchDB.py#L226-L227

aldbr: @bertrandrigaud please choose one:

1. thisisprefix
2. thisisprefix_

index


thisisprefix_index

aldbr: 2 is more explicit - @vkuznet

sfayer: Hi,

We'd prefer option 2 as it's more flexible: Our global prefix separator is "." on our instance (although we're also the admins of that so could change it if really needed).

Regards,
Simon

bertrandrigaud: OK to me, global prefix should be declared in values.yaml like this:

**GlobalIndexPrefix: myprefix**

So that in Opensearch we have for example:

**myprefix_job_parameters_dirac_0m**

But I understand Simon's point of view, so in this case it would be **myprefix._job_parameters_dirac_0m** which is not pretty. So for this reason, let's add "_" to the GlobalIndexPrefix variable and not in the code:

**GlobalIndexPrefix: myprefix_**

chaen: Good with me, I implement the change

aldbr (approved):

PR #929, opened 2026-06-02
fix: clean up AuthDB refresh tokens by dropping partitions

### LINKED ISSUE #928: [BUG]: Refresh tokens cleaned up too slowly
In https://github.com/DIRACGrid/diracx-charts/issues/222 refresh token clean up was intended to be implemented by calling TRUNCATE on partitions which are too old.

Instead old refresh tokens are being cleaned with a simple DELETE which has two issues:

* A huge number of rows are deleted in each query, resulting in a massive undo log and poor performance
* Newer partitions aren't being maintained

Instead we should:

* [x] Call TRUNCATE on old partitions and drop them
* [x] Create new partitions if the next two years worth of partitions don't exist
* [ ] Add clean up to the DeviceFlows and AuthorizationFlows tables

chrisburr:
Replace the row-level DELETE sweeps for the RefreshTokens table with maintenance of its monthly JTI range-partitions: drop partitions whose whole month is older than the retention horizon, and add partitions ahead of time so the p_future catch-all never fills. Dropping a partition is an O(1) metadata operation and avoids the row-lock and lock-memory cost of large DELETEs.

Retention is now expressed in calendar months via the new `DIRACX_SERVICE_AUTH_REFRESH_TOKEN_RETENTION_MONTHS` setting (default 6), replacing revoked_refresh_token_retention_minutes. Partition maintenance is implemented for MySQL only and raises NotImplementedError for other dialects. The unpartitioned flow tables keep their existing DELETE-based cleanup.

Fixes #928


aldbr (approved):

diracx-db/src/diracx/db/sql/auth/db.py
```diff
@@ -340,37 +394,84 @@ async def revoke_user_refresh_tokens(self, subject):
...
 
-        return res_revoked.rowcount
+        # Add new partitions first, by splitting the p_future catch-all.
+        if months_to_add:
+            new_partitions = [
+                f"PARTITION {_partition_name(m)} "
+                f"VALUES LESS THAN ('{_partition_boundary(m + relativedelta(months=1))}')"
+                for m in months_to_add
+            ]
+            new_partitions.append("PARTITION p_future VALUES LESS THAN (MAXVALUE)")
```
chaen, line 454: we are already creating partitions in the post-update, isn't that a duplicate ?

chrisburr: From chatting here, yes I'd missed that. Though at least LHCb doesn't have init-sql enabled at the moment so it's probably safer having it here and mostly harmless having it in two places.

PR #931, opened 2026-06-04
docs: document how to test diracx-charts changes

chaen:


docs/dev/how-to/test-changes-in-diracx-charts.md
fstagni, line 1: Shouldn't this page be linked from somewhere? Otherwise it will not be built.

docs/dev/how-to/test-changes-in-diracx-charts.md
```diff
@@ -0,0 +1,26 @@
+# Testing changes in diracx-charts
+
+Changes in `diracx` may require corresponding changes in `diracx-charts`.
```
fstagni: Add a "typical case", an example of when it is needed.

docs/dev/how-to/test-changes-in-diracx-charts.md
```diff
@@ -0,0 +1,26 @@
+# Testing changes in diracx-charts
```
fstagni:
```suggestion
# Testing against changes in diracx-charts
```

fstagni: BTW, this is valid also for diracx-web, no?

aldbr: It's true but may be a bit less important than `diracx` as we rarely need to modify `diracx-charts` when modifying `diracx-web` (e.g. no DB to add to the charts)

fstagni: I was more meaning that if you need to test `diracx` together with `diracx-web` is the same technique?

aldbr: Indeed, but theoretically `diracx-web` should only be built on top of stable `diracx` endpoints. A developer that would want to introduce a new endpoint would first need to get it in `diracx` validated, merged and released before working on the `diracx-web` interface (may be this should be documented though).

chaen: Could you add that in a `diracx-web` doc ?

fstagni (approved):

PR #910, opened 2026-05-06
feat: add a Resource status source and route

Loxeris:
Closes #839.

The work is not yet finished and tests are currently failing.

The most important problem is detailed [in this comment](https://github.com/DIRACGrid/diracx/issues/839#issuecomment-4388382820) regarding the use of `CacheableSource` and sync/async incompatibilities.

The sources should also be more likely instantiated in `factory.py` and router would get values through dependency injection.

PR #936, opened 2026-06-11
feat(core): add AsyncTwoLevelCache

### LINKED ISSUE #839: ResourceStatusSource + route
### Description

Wire the data models https://github.com/DIRACGrid/diracx/issues/837 and caching infrastructure https://github.com/DIRACGrid/diracx/issues/838 together: implement `ResourceStatusSource(CacheableSource)` that reads from `ResourceStatusDB` and serves resource status through a new config route.

### Scope

1. **`ResourceStatusSource`**: a `CacheableSource` subclass that:
   - Reads from `ResourceStatusDB` (Site + Resource status tables)
   - Translates DB status values (`Active`/`Degraded`/`Banned`/...) to the binary Pydantic models (`AllowedStatus`/`BannedStatus`)
   - Implements `latest_revision()` — e.g. hash of current status snapshot, or max `DateEffective`/`LastCheckTime`
   - Implements `read_raw()` — returns the full status dict per resource type and VO

2. **Route**: new endpoint under the config router:
   ```
   /config/resource-status/storage?vo={inferred_from_credentials} -> dict[str, StorageElementStatus]
   ```
   Equivalent endpoints for compute and site status.

3. **Access policy**: any authenticated user can read (similar to current config endpoint being `@open_access`, but scoped to the user's VO).

4. **Integration tests**: verify end-to-end flow from DB data to API response, including HTTP caching headers (ETag / 304 Not Modified).

### Technical details

- `ResourceStatusSource` lives in `diracx-core/src/diracx/core/config/` or a new `diracx-core/src/diracx/core/rss/` package
- VO is inferred from the client's token — no need to expose all VOs
- HTTP caching (ETag, Last-Modified, 304) should reuse the same pattern as the config endpoint
- Dependency injection: `ResourceStatusSource` needs access to `ResourceStatusDB` — design the wiring so it fits the existing DI patterns

### Reference

- Chris's route proposal: [#790 comment](https://github.com/DIRACGrid/diracx/issues/790#issuecomment-4012486144)
- Config router pattern: `diracx-routers/src/diracx/routers/configuration.py`
- DI patterns: `diracx-routers/src/diracx/routers/dependencies.py`

### Acceptance criteria

- [ ] `ResourceStatusSource(CacheableSource)` reads from `ResourceStatusDB` and produces Pydantic status models
- [ ] Route serves storage/compute/site status per VO
- [ ] HTTP caching headers (ETag, 304 Not Modified)
- [ ] Access policy: authenticated users, scoped to their VO
- [ ] Integration tests covering the full read path

chrisburr:
Async equivalent of TwoLevelCache for use with coroutine populate functions: same soft/hard TTL semantics, but coordinated with asyncio primitives instead of a ThreadPoolExecutor.

Replaces #910
Part of #839

### DIFF
--- a/diracx-core/src/diracx/core/utils.py
+++ b/diracx-core/src/diracx/core/utils.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 __all__ = [
     "EXPIRES_GRACE_SECONDS",
+    "AsyncTwoLevelCache",
     "TwoLevelCache",
     "batched_async",
     "dotenv_files_from_environment",
@@ -11,6 +12,7 @@ __all__ = [
     "write_credentials",
 ]
 
+import asyncio
 import fcntl
 import json
 import logging
@@ -19,7 +21,7 @@ import re
 import stat
 import threading
 from collections import defaultdict
-from collections.abc import Callable
+from collections.abc import Callable, Coroutine
 from concurrent.futures import Future, ThreadPoolExecutor
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


diracx-core/src/diracx/core/utils.py
```diff
@@ -293,6 +295,172 @@ def clear(self):
...
+                self.tasks[key] = task
+            task = self.tasks[key]
+
+            if key in self.hard_cache:
+                # Soft miss but hard hit: serve stale while the refresh runs.
+                # Pre-fill soft cache so the next request skips the lock entirely.
+                result = self.hard_cache[key]
+                self.soft_cache[key] = result
+                return result
+
```
Loxeris, line 397: Is the soft-cache pre-filling necessary ? This isn't a big problem I think but I imagine this mechanism could technically extend the hard TTL a bit.

chrisburr: Thanks for reviewing!

It's an optimisation: without it, every request during the refresh window would have to acquire the (single, shared) lock instead of taking the lock-free fast path. Acquiring locks is relatively expensive and is by definition sequential so it could massively harm DiracX performance.

So long as soft-TTL << hard-TTL the effect should be small and it only materialises if `_work` fails or is still running when the hard entry expires.

(I'm basing this on "vibes" so feel free to argue against me on this or any other comment I make if my justification seems to not make sense.)

Loxeris: Alright, this optimisation makes sense then. I just wasn't sure about how performance expensive the part above was.

Loxeris (approved):
Seems good to me!

PR #945, opened 2026-06-12
docs: add explanation page on splitting pull requests

chrisburr:


fstagni:
While I see the purpose, it feels like this is more of a blog post rather than something that should go to documentation.

aldbr:
> While I see the purpose, it feels like this is more of a blog post rather than something that should go to documentation.

What do you mean? Do we have a DiracX blog? :sweat_smile: 

It is directly linked to the contribution guidelines, which appear in every DIRACGrid repositories under the "Contributing" tab, so I think it's the right place.
image

Do you have any other suggestion then?

fstagni:
OOk

fstagni (approved):

PR #950, opened 2026-06-16
fix: added policies to RefreshToken DB

### LINKED ISSUE #524: extra token information coming from policies are not persisted in the refreshtoken DB

Stellatsuu:
### Issue found:
Policies were added by `mint_token` after `insert_token` (in DB) before `exchange_token` was called, so they were never added to the DB, only to the JWT. `RefreshTokens` DB was also missing a Policy column.

### Changes:
- `RefreshTokens` now has a `policies` column (updates to the DB tablecolumns might be needed)
- Policies are now added before inserting the token in the DB in the `exchange_token` function.
- `mint_token` function doesn't need to enrich the token anymore since the payloads already have the policies inside them, from `exchange_token` function.
- Removed unnecessary arguments from `enrich_token` function (?). I'm not sure if they are really unnecessary -> see about the lack of usage of `enrich_token`, : https://github.com/DIRACGrid/diracx/pull/950#issuecomment-4717200093

- In the DB, `RefreshToken` table: 
```
| 019ed007e31a7ec0980b77d7bf880d28 | CREATED | vo:diracAdmin | diracAdmin:CiRBRkFFNjIzOS0wRkIzLTRFQUEtQUJGRC0zQ0MxRDQ1MzJBNDESBWxvY2Fs | {"wms": {"PolicySpecific": "TemporaryRefreshPolicy"}, "sandbox": {"PolicySpecific": "TemporaryRefreshPolicy"}} |
```
- In the JWT, refresh token payload: 
```
{
  "jti": "019ed007-e31a-7ec0-980b-77d7bf880d28",
  "exp": 1781628238,
  "dirac_policies": {
    "sandbox": {
      "PolicySpecific": "TemporaryRefreshPolicy"
    },
    "wms": {
      "PolicySpecific": "TemporaryRefreshPolicy"
    }
  },
  "legacy_exchange": false
} 
```

### DIFF
--- a/diracx-db/src/diracx/db/sql/auth/schema.py
+++ b/diracx-db/src/diracx/db/sql/auth/schema.py
@@ -119,5 +119,6 @@ class RefreshTokens(Base):
 
     # User attributes bound to the refresh token
     sub: Mapped[str] = mapped_column("Sub", String(256), index=True)
+    policies: Mapped[dict[str, Any]] = mapped_column("Policies", index=True)
 
     __table_args__ = (Index("index_status_sub", status, sub),)
--- a/diracx-db/src/diracx/db/sql/auth/schema.py
+++ b/diracx-db/src/diracx/db/sql/auth/schema.py
@@ -116,9 +116,9 @@ class RefreshTokens(Base):
         "Status", RefreshTokenStatus, server_default=RefreshTokenStatus.CREATED.name
     )
     scope: Mapped[str1024] = mapped_column("Scope")
+    policies: Mapped[dict[str, Any]] = mapped_column("Policies")
 
     # User attributes bound to the refresh token
     sub: Mapped[str] = mapped_column("Sub", String(256), index=True)
-    policies: Mapped[dict[str, Any]] = mapped_column("Policies", index=True)
 
     __table_args__ = (Index("index_status_sub", status, sub),)
--- a/diracx-logic/src/diracx/logic/auth/token.py
+++ b/diracx-logic/src/diracx/logic/auth/token.py
@@ -4,9 +4,10 @@ from __future__ import annotations
 
 import base64
 import hashlib
+import logging
 import re
 from datetime import datetime, timedelta, timezone
-from typing import cast
+from typing import Any, cast
 
 from joserfc import jwt
 from joserfc.jwt import Claims
@@ -36,6 +37,8 @@ from .utils import (
     parse_and_validate_scope,
     verify_dirac_refresh_token,
 )
+
+logger = logging.getLogger(__name__)
 
 
 async def get_oidc_token(
@@ -317,13 +320,12 @@ async def exchange_token(
     sub = f"{vo}:{sub}"
 
     refresh_payload: RefreshTokenPayload | None = None
+
     if include_refresh_token:
         # Insert the refresh token with user details into the RefreshTokens table
         # User details are needed to regenerate access tokens later
         refresh_jti = await insert_refresh_token(
-            auth_db=auth_db,
-            subject=sub,
-            scope=scope,
+            auth_db=auth_db, subject=sub, scope=scope, policies={}
         )
 
         # Generate refresh token payload
@@ -391,9 +393,7 @@ def _sign_token_payload(claims: dict, settings: AuthSettings) -> str:
 
 
 async def insert_refresh_token(
-    auth_db: AuthDB,
-    subject: str,
-    scope: str,
+    auth_db: AuthDB, subject: str, scope: str, policies: dict[str, Any]
 ) -> UUID:
     """Insert a refresh token into the database and return the JWT ID."""
     # Generate a JWT ID
@@ -404,6 +404,7 @@ async def insert_refresh_token(
         jti=jti,
         subject=subject,
         scope=scope,
+        policies=policies,
     )
     return jti
 
--- a/diracx-logic/src/diracx/logic/auth/token.py
+++ b/diracx-logic/src/diracx/logic/auth/token.py
@@ -48,6 +48,7 @@ async def get_oidc_token(
     config: Config,
     settings: AuthSettings,
     available_properties: set[SecurityProperty],
+    policies: dict[str, Any],
     device_code: str | None = None,
     code: str | None = None,
     redirect_uri: str | None = None,
@@ -90,6 +91,7 @@ async def get_oidc_token(
     return await exchange_token(
         auth_db,
         scope,
+        policies,
         oidc_token_info,
         config,
         settings,
@@ -238,6 +240,7 @@ async def perform_legacy_exchange(
     expected_api_key: str,
     preferred_username: str,
     scope: str,
+    policies: dict[str, Any],
     authorization: str,
     auth_db: AuthDB,
     available_properties: set[SecurityProperty],
@@ -264,6 +267,7 @@ async def perform_legacy_exchange(
     return await exchange_token(
         auth_db,
         scope,
+        policies,
         {"sub": sub, "preferred_username": preferred_username},
         config,
         settings,
@@ -276,6 +280,7 @@ async def perform_legacy_exchange(
 async def exchange_token(
     auth_db: AuthDB,
     scope: str,
+    policies: dict[str, Any],
     oidc_token_info: dict,
     config: Config,
     settings: AuthSettings,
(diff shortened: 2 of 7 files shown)


Stellatsuu:
@aldbr I have a question:

 Is it normal that none of the policies (`WMSAccessPolicy`, `SandboxAccessPolicy`, `LollygagAccessPolicy` and `MyPilotsAccessPolicy`) implements this function?:
https://github.com/DIRACGrid/diracx/blob/ebce8adeef4ce736fdba1b0dab0f9e9ee7867a27/diracx-routers/src/diracx/routers/access_policies.py#L93-L104

The only one having it is `AlwaysAllowAccessPolicy`, used for the tests. Even by adding the policies to the DB, they would only be empty dicts without this function implemented no?

Since all the work for adding policies to token is done here:
https://github.com/DIRACGrid/diracx/blob/ebce8adeef4ce736fdba1b0dab0f9e9ee7867a27/diracx-routers/src/diracx/routers/auth/token.py#L56-L76

chrisburr:
> Is it normal that none of the policies (WMSAccessPolicy, SandboxAccessPolicy, LollygagAccessPolicy and MyPilotsAccessPolicy) implements this function?:

Yes, this is there as something that we know we will need but that we don't yet have a use for.

diracx-routers/src/diracx/routers/access_policies.py
```diff
@@ -90,15 +86,10 @@ async def policy(policy_name: str, user_info: AuthorizedUserInfo, /):
         return
 
     @staticmethod
-    def enrich_tokens(
-        access_payload: AccessTokenPayload, refresh_payload: RefreshTokenPayload | None
-    ) -> tuple[dict, dict]:
+    def enrich_tokens() -> tuple[dict, dict]:
```
aldbr, line 89: Any reason for modifying the `enrich_tokens` method?

diracx-logic/src/diracx/logic/auth/token.py
```diff
@@ -37,6 +38,8 @@
     verify_dirac_refresh_token,
 )
 
+logger = logging.getLogger(__name__)
```
aldbr, line 41: I see your `logger`, which is likely fine, but I don't see where you are using it

diracx-logic/src/diracx/logic/auth/token.py
```diff
@@ -316,14 +324,23 @@ async def exchange_token(
...
 
+    # Enrich the token with policy specific content
+    dirac_access_policies = {}
+    dirac_refresh_policies = {}
+    for policy_name, policy in policies.items():
+        access_extra, refresh_extra = policy.enrich_tokens()
+        if access_extra:
+            dirac_access_policies[policy_name] = access_extra
+        if refresh_extra:
+            dirac_refresh_policies[policy_name] = refresh_extra
```
aldbr, line 335: When you renew the refresh token, you don't reuse the policies that you initially stored in the DB, is that expected? (then what's the point in storing the policies in the DB?)

What if the policies changed in the meantime? Then you get a different refresh token.
Please test this use case if possible

diracx-db/src/diracx/db/sql/auth/schema.py
```diff
@@ -116,6 +116,7 @@ class RefreshTokens(Base):
         "Status", RefreshTokenStatus, server_default=RefreshTokenStatus.CREATED.name
     )
     scope: Mapped[str1024] = mapped_column("Scope")
+    policies: Mapped[dict[str, Any]] = mapped_column("Policies")
```
aldbr, line 119: I think it should be `null` by default, at least for the migration (because old refresh tokens don't have `Policies`. Also it would probably by `null` permanently, am I wrong?

aldbr (changes requested):
Can you also design a migration plan for the installation already in production please (and make sure the code is able to handle that)? 🙂 

Thanks!

diracx-logic/src/diracx/logic/auth/token.py
Stellatsuu: I was testing something there, I thought I removed it.

diracx-db/src/diracx/db/sql/auth/schema.py
Stellatsuu: I agree about the `null` by default and I think it's automatically assigned for older values when you add the column in the DB with `ALTER TABLE RefreshTokens ADD COLUMN Policies JSON;` (`sql-alchemy` only auto-check for new tables, not new columns, I don't know if there's a setting for that), at least, it did it for my older tokens. I will re-check that if you want and add a default anyway.

I don't understand that:
> Also it would probably by null permanently, am I wrong?

As I mentionned in the PR, the policies are stored correctly in the DB, it's not `null`:
> made with a temporary change in `BaseAccessPolicy.enrich_token` to return `{}, {"TemporaryRefreshPolicy"}`
> - In the DB, `RefreshToken` table: 
```
| 019ed007e31a7ec0980b77d7bf880d28 | CREATED | vo:diracAdmin | diracAdmin:CiRBRkFFNjIzOS0wRkIzLTRFQUEtQUJGRC0zQ0MxRDQ1MzJBNDESBWxvY2Fs | {"wms": {"PolicySpecific": "TemporaryRefreshPolicy"}, "sandbox": {"PolicySpecific": "TemporaryRefreshPolicy"}} |
```

If you're talking about this case: https://github.com/DIRACGrid/diracx/pull/950#discussion_r3458551398, I will take a look at it.

diracx-routers/src/diracx/routers/access_policies.py
Stellatsuu: These changes were made due to my comment here: https://github.com/DIRACGrid/diracx/pull/950#issuecomment-4717200093

I wasn't sure `access_payload` and `refresh_payload` had an impact or not since there was no real implementation of this method.

Also, if we need the payload to enrich the token, how can we do that? Because currently, the payload is created after we insert the token in the DB, and so, after we enrich it too:

```python
# Enrich the token with policy specific content 
    dirac_access_policies = {}
    dirac_refresh_policies = {}
    for policy_name, policy in policies.items():
        access_extra, refresh_extra = policy.enrich_tokens() # --> with my refactor, token is enriched first here
        if access_extra:
            dirac_access_policies[policy_name] = access_extra
        if refresh_extra:
            dirac_refresh_policies[policy_name] = refresh_extra

    refresh_payload: RefreshTokenPayload | None = None

    if include_refresh_token:
        # Insert the refresh token with user details into the RefreshTokens table
        # User details are needed to regenerate access tokens later
        refresh_jti = await insert_refresh_token(
            auth_db=auth_db, subject=sub, scope=scope, policies=dirac_refresh_policies
        ) # --> we need to retrieve the JTI here

        # Generate refresh token payload
        if refresh_token_expire_minutes is None:
            refresh_token_expire_minutes = settings.refresh_token_expire_minutes
        refresh_exp = uuid7_to_datetime(refresh_jti) + timedelta(
            minutes=refresh_token_expire_minutes
        )
        refresh_payload = RefreshTokenPayload(
            jti=str(refresh_jti),
            exp=refresh_exp,
            # legacy_exchange is used to indicate that the original refresh token
            # was obtained from the legacy_exchange endpoint
            legacy_exchange=legacy_exchange,
            dirac_policies=dirac_refresh_policies,
        ) # --> the payload is only created here
``` 

Maybe one way to do that would be to create an "update_token_policies" function and call it at the end, so it would be:
- create refresh token in DB
- create refresh and access tokens payloads
- enrich them
- update refresh token and access payloads dirac_policies
- update refresh token policies in DB with function

What do you think? Do you have any other idea?

diracx-logic/src/diracx/logic/auth/token.py
Stellatsuu: > When you renew the refresh token, you don't reuse the policies that you initially stored in the DB, is that expected?  (then what's the point in storing the policies in the DB?)

I think so, no? The issue was asking to add the policies to the DB, since they were only JWT-side, it's true that I don't see the point in storing them in DB if we don't reuse them but I also don't see a point of reusing them if the refresh token will always be re-created by calling `exchange_token`.

> What if the policies changed in the meantime? Then you get a different refresh token.

Since we don't reuse the policies from the DB and the token is re-created anyway, you'll have a new token with the new policies?

aldbr:
> ddev:
- extra field in the access token, any policy can add content to that (not persisted when the refresh token is renewed)
- a bit of doc in the acces policy base

> blocking PR for now: @chaen and I are going to go through it and will come up with a better description.

Stellatsuu:
Closed due to changes of plans: https://github.com/DIRACGrid/diracx/issues/524#issuecomment-4395561176

Will create another MR.

Issue #914, opened 2026-05-11
Harden GitHub actions configuration


chrisburr:
It would be good to review https://cheatsheetseries.owasp.org/cheatsheets/GitHub_Actions_Security_Cheat_Sheet.html and make sure we're following all of the best practices.

This issue is also a good chance for someone to learn about all of the various security foot-guns of GHA.

aldbr:
> ddev: @chaen will add a link about a potential official Github page related to that too

aldbr:
> ddev: if any fuzzy issue is encountered along the road, then we will open a dedicated issue to solve it

chrisburr:
I've been recommended this tool: https://github.com/zizmorcore/zizmor

PR #954, opened 2026-06-18
chore: supply chain security improvements

chrisburr:


aldbr:
Just wondering, is that part of https://github.com/DIRACGrid/diracx/issues/914?

chrisburr:
Maybe though I expect that to flag other things and would be surprised if it knew about pixi's dependency cooldowns.

I mostly made this PR as I thought we already had them configured.

aldbr (approved):
LGTM

PR #956, opened 2026-06-23
ci: harden github actions

### LINKED ISSUE #914: Harden GitHub actions configuration

Stellatsuu:
### Changes
- Applied [Zizmor](https://github.com/zizmorcore/zizmor) recommandations in GHA files
- Applied [Pinact](https://github.com/suzuki-shunsuke/pinact) to update checkout versions in GHA files, recommended by Zizmor (https://docs.zizmor.sh/audits/#ref-version-mismatch)
- Added Zizmor to pre-commit
- Added Zizmor and Pinact tasks in pixi: devs can use `pixi run zizmor --fix=safe` to auto-fix minor issues or `pixi run pinact` to auto-update checkout versions

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -12,6 +12,8 @@ on:
   schedule:
     - cron: '0 10 * * *'
 
+permissions: {}
+
 jobs:
   deploy-pypi:
     name: PyPI deployment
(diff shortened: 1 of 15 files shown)


chrisburr (approved):
Thanks for this! 🍰

PR #957, opened 2026-06-23
test: replace sleeps with freezegun time ticks in tests

### LINKED ISSUE #519: Don't use time.sleep in tests to avoid unique constraint exceptions
Think of other ways to advance time where required e.g.
- where multiple inserts are required with a `DATETIME('now')` column and there is a unique constaint on that column.

Possible ways to achieve this:
- use freezegun to mock datetime.now and advance time as needed inside a test between inserts. overwrite the datetime functions in sqlite as required. Use a pragma in the ClientFactory with an on connect listener to overwrite the functions on each sqlite connection used.

Problems encountered on first attempts:

- Freezegun needed to be initialised quite early on to ensure datetime was mocked before initialising the app and ClientFactory
- When not using freezegun, typing issues e.g. when serializing fake datetime objects

Discussed in

https://github.com/DIRACGrid/diracx/pull/516#discussion_r2076915044_

bndikt:
## Summary
- replace wall-clock sleeps in tests with `freezegun` time ticks
- wire SQLite test connections to `mock_sqlite_time` so `DATETIME('now')` follows frozen Python time
- freeze tests at the current UTC second to match SQLite timestamp precision

### DIFF
--- a/diracx-db/tests/jobs/test_sandbox_metadata.py
+++ b/diracx-db/tests/jobs/test_sandbox_metadata.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-import asyncio
 import secrets
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import pytest
 import sqlalchemy
@@ -11,12 +10,14 @@ from diracx.core.exceptions import SandboxAlreadyInsertedError, SandboxNotFoundE
 from diracx.core.models import SandboxInfo, UserInfo
 from diracx.db.sql.sandbox_metadata.db import SandboxMetadataDB
 from diracx.db.sql.sandbox_metadata.schema import SandBoxes, SBEntityMapping
+from diracx.testing.time import install_sqlite_time_mock
 
 
 @pytest.fixture
 async def sandbox_metadata_db(tmp_path):
     sandbox_metadata_db = SandboxMetadataDB("sqlite+aiosqlite:///:memory:")
     async with sandbox_metadata_db.engine_context():
+        install_sqlite_time_mock(sandbox_metadata_db.engine)
         async with sandbox_metadata_db.engine.begin() as conn:
             await conn.run_sync(sandbox_metadata_db.metadata.create_all)
         yield sandbox_metadata_db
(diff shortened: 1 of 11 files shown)


diracx-db/tests/jobs/test_sandbox_metadata.py
```diff
@@ -1,27 +1,39 @@
...
     async with sandbox_metadata_db.engine_context():
+        sqlalchemy.event.listen(
+            sandbox_metadata_db.engine.sync_engine, "connect", mock_sqlite_time
+        )
         async with sandbox_metadata_db.engine.begin() as conn:
             await conn.run_sync(sandbox_metadata_db.metadata.create_all)
         yield sandbox_metadata_db
 
 
+@pytest.fixture
```
chrisburr: As this is defined in two places it probably belongs in `diracx-testing/src/diracx/testing/time.py`

diracx-testing/src/diracx/testing/utils.py
```diff
@@ -335,6 +336,9 @@ def set_sqlite_pragma(dbapi_connection, connection_record):
                 sqlalchemy.event.listen(
                     db.engine.sync_engine, "connect", set_sqlite_pragma
                 )
+                sqlalchemy.event.listen(
+                    db.engine.sync_engine, "connect", mock_sqlite_time
```
chrisburr: Similarly, this is now in three places, maybe it can be simplified with a helper:

```python
# diracx-testing/src/diracx/testing/time.py
def install_sqlite_time_mock(engine) -> None:
    """Make SQLite's date/time functions follow frozen Python time on `engine`.

    Registers `mock_sqlite_time` as a "connect" listener so every new DBAPI
    connection gets the mocked DATETIME()/JULIANDAY()/... functions. Accepts a
    sync Engine or an AsyncEngine.
    """
    sync_engine = getattr(engine, "sync_engine", engine)
    sqlalchemy.event.listen(sync_engine, "connect", mock_sqlite_time)

# utils.py:339  (create_db_schemas)
install_sqlite_time_mock(db.engine)

# test_sandbox_metadata.py:21  (sandbox_metadata_db fixture)
install_sqlite_time_mock(sandbox_metadata_db.engine)

# test_freeze_time.py:43
if with_mock:
    install_sqlite_time_mock(engine)
```

chrisburr (dismissed):

aldbr:
@chrisburr I noticed `freezegun` and `sqlalchemy` are used within `diracx-testing` but not part of the `pyproject.toml`. I assume this was forgotten, unless I miss something here?

aldbr:
> ddev: usually we add the import at the top of the modules, we should probably add an explicit rule about that (guidelines for devs)

aldbr (approved):

diracx-db/tests/jobs/test_sandbox_metadata.py
bndikt: Good call. Created two helper functions in `diracx-testing/src/diracx/testing/time.py` as suggested.

PR #967, opened 2026-07-02
chore: token policies persisted in DB

### LINKED ISSUE #524: extra token information coming from policies are not persisted in the refreshtoken DB

Stellatsuu:
### Changes:

- Moved the `dirac_policies` from `TokenPayload` to `RefreshTokensPayload` and  `AccessTokenPayload`. 
**Later, this should be removed from `RefreshTokensPayload`**.

- `RefreshTokens` now has a `policies` column. 
**Manual updates to the already existing DBTable is needed: `ALTER TABLE RefreshTokens ADD COLUMN Policies JSON`. It should add NULL by default for existing values.**

- `Policies` are added from `all_access_policies` in the `exchange_token` function if no policies were found in the DB (e.g, new `RefreshToken`). Otherwise, they are taken from the DB to create a new `AccessTokenPayload`.

- `mint_token` function doesn't need to enrich the token anymore since the payloads already have the policies inside them, from `exchange_token` function.

- `enrich_token` function now takes only the `AccessTokenPayload` and returns one dict with the policies values.

### DIFF
--- a/diracx-client/tests/test_auth.py
+++ b/diracx-client/tests/test_auth.py
@@ -17,7 +17,6 @@ REFRESH_CONTENT = {
     "jti": "f0706e0a-af1e-4538-9f1f-7b9620783cba",
     "exp": int((datetime.now(tz=timezone.utc) + timedelta(days=1)).timestamp()),
     "legacy_exchange": False,
-    "dirac_policies": {},
 }
 
 TOKEN_RESPONSE_DICT = {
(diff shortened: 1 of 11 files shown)


fstagni:
Hi, can you resolve the conflicts?

aldbr:
Hi @Stellatsuu, thanks for the fast update :slightly_smiling_face: 

We have discussed it among the maintainers and we actually would like to hold that PR for now. As it includes DB changes, we see it as a good opportunity to leverage Alembic first.

We currently have no migration mechanism and we are close to have multiple installations running diracx (e.g. you are already running it in certification) so it would be easier to have Alembic doing the job.

I will open a new issue for Alembic and your work will be held until it's done, and you will just have to rebase your PR and add the alembic script for the migration basically.

What do you think?

Stellatsuu:
Hi @aldbr, no problem for me :)

PR #978, opened 2026-07-08
feat: add RSS read adapter for DIRAC

Loxeris:
closes DIRACGrid/DIRAC#8685

- new self-contained module dedicated to the rss read adapter for Dirac.


Loxeris:
A few points/issues about the current state:

- [ResourceStatus](https://github.com/DIRACGrid/DIRAC/blob/e2586a6f8526bb8d1221eba12522a1173df18e2c/src/DIRAC/ResourceStatusSystem/Client/ResourceStatus.py#L127) expects VO in the data it uses to refresh the cache, but it's not provided by the RSS API, should I use a default value (currently None) or make changes to the RSS API output data ?
- The Pydantic model uses `"Unknown"` as the default reason for banned statuses, but it's also used for the `Unknown` statuses. For now I assume it's a `Banned` status as it seems more frequent.
- There is currently no way to specify we want all the VOs from the RSS API as it is inferred from the credentials.

aldbr (changes requested):
Oops my bad, this layer should be integrated to DIRAC, not diracx.
I added the issue to the diracx repo but it was a mistake, and I just moved it to DIRAC.

I am going to close this PR, please reopen it in DIRAC, you will have to make a few adjustment:
- you will have to use the `diracx-client` to call the `diracx-routers` from DIRAC, through the `Impersonator`. See how it's done in `SandboxStoreHandler` in DIRAC.
- Also we don't want to get back `Degraded` and `Probing` because DIRAC is going to be streamlined to just contain `Active` and `Banned` in most of the cases normally (see first issue of Phase2)

PR #980, opened 2026-07-08
fix: from minio to seaweedfs

### LINKED ISSUE #191: Replace minio with something else
Background: https://github.com/minio/minio/issues/21647#issuecomment-3418675115

🙄 

Let's see how it settles. Shouldn't affect any prod deployments as most people are better off using pre-existing S3-storage or an S3 plugin for their existing storage (like the CEPH plugin).

bertrandrigaud:
Move from MinIO to Seaweedfs. This should unlock s3 testing part
Modify test_s3.py in order to fit diracx-charts new seaweedfs config (values.yaml)
With Minio, the checksum test was waiting for a clear error word "ContentChecksumMismatch" which is not the case for seaweedfs where we have now a sentence like "The Content-Md5 you specified did not match what we received"
So for the moment the test is expected this exact sentence, maybe it's too long...? Tell me.
We have to pay attention in case this sentence changes in the future releases of seaweedfs and also if we need to move again to another s3 software

### DIFF
--- a/diracx-cli/src/diracx/cli/internal/legacy.py
+++ b/diracx-cli/src/diracx/cli/internal/legacy.py
@@ -197,7 +197,7 @@ def generate_helm_values(
         "initSql": {"enabled": False},
         "cert-manager": {"enabled": False},
         "cert-manager-issuer": {"enabled": False},
-        "minio": {"enabled": False},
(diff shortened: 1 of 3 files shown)


aldbr:
Duplicate of https://github.com/DIRACGrid/diracx/pull/981

PR #988, opened 2026-07-10
refactor(routers): remove DiracHttpResponseError and simplify OAuth pending response handling

### LINKED ISSUE #576: `DiracError`, do we keep it?
`DiracError` adds http error support in the factory. Do we still need it?

```py
class DiracError(RuntimeError):
    http_status_code = HTTPStatus.BAD_REQUEST  # 400
    http_headers: dict[str, str] | None = None

    def __init__(self, detail: str = "Unknown"):
        self.detail = detail
```

In the factory we have the following part:

```py
def dirac_error_handler(request: Request, exc: DiracError) -> Response:
    return JSONResponse(
        status_code=exc.http_status_code,
        content={"detail": exc.detail},
        headers=exc.http_headers,
    )
```

...which is useful. But the problem I see is that no error is using that feature, and we automatically catch errors to later throw the intended error (`try... except JobNotFoundError: throw...`).

Catching errors explicitly with `try... catch` is useful to make the code more readable (we can see what behavior we are waiting for).

> So my question is, do we keep it?

I would even prefer that by default, if an error is not caught, we raise an 5XX error because it was not expected.

mmascher:
## Summary
This PR updates the auth error-handling approach in routers by removing DiracHttpResponseError and handling the OAuth pending-authorization response directly in the token route.

## What changed
- Removed the custom DiracHttpResponseError path from routers.
- Returned the OAuth authorization_pending payload directly from the token endpoint using a JSON response.
- Removed now-unneeded exception handler wiring in app factory.
- Updated coding conventions docs to reflect the current approach:
  - protocol-specific payloads can be returned directly when HTTPException shape is not suitable.

## Why
The custom exception class became a one-off indirection with little value:
- direct response in the endpoint is clearer,
- behavior is easier to follow in reviews,
- less plumbing in global exception handling.


docs/dev/reference/coding-conventions.md
```diff
@@ -135,3 +135,10 @@ To make sure that each part of DiracX is doing only what it is supposed to do, y
...
 - `diracx-logic` should embed Dirac specificities. It should encapsulate the logic of the services and should call `diracx-db` to interact with databases.
 - `diracx-db` should contain atomic methods (complex logic is expected to be located in `diracx-db`).
+
+### Error handling boundaries
+
+- `diracx-db` and `diracx-logic` should raise domain exceptions (typically `DiracError` subclasses), and should not depend on `fastapi`.
+- `diracx-routers` is the HTTP boundary and should translate known backend/domain exceptions into `HTTPException` with the desired status code, message, and headers.
+- `DiracError` handling registered at app level (see [diracx-routers/src/diracx/routers/factory.py](../../../diracx-routers/src/diracx/routers/factory.py), especially `app.add_exception_handler(DiracError, cast(handler_signature, dirac_error_handler))`) is a fallback safety net for uncaught domain exceptions, not a replacement for explicit mapping when the router needs specific HTTP behavior.
+- Protocol-specific responses can be handled directly in routers when `HTTPException` shape is not suitable (for example OAuth2-style payloads). In that case DiracHttpResponseError shall be used.
```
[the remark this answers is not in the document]
mmascher: Godd catch bot, I forgot about that!

.gitignore
```diff
@@ -112,3 +112,4 @@ site
 # DiracX specific
 # No point in committing the pixi.lock for gubbins as it cannot work properly within the diracx repo
 extensions/gubbins/pixi.lock
+TIPS
```
aldbr: Oops I assume it's unexpected? 😅 
```suggestion
```

diracx-routers/src/diracx/routers/exceptions.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+__all__ = ["DiracHttpResponseError"]
+
+
+class DiracHttpResponseError(RuntimeError):
```
aldbr: This one is just used once in a very specific context because the body of the device flow polling response must contain an `error` key instead of a `detail` one provided by `HTTPException`. 
Since it's very specific and only used once, I think I would just remove it and use `JSONResponse` directly in `diracx-routers` `token.py`:

```python
return JSONResponse(
    status_code=HTTPStatus.BAD_REQUEST,
    content={"error": "authorization_pending"},
)
```

Any opinion?

docs/dev/reference/coding-conventions.md
```diff
@@ -135,3 +135,10 @@ To make sure that each part of DiracX is doing only what it is supposed to do, y
...
 - `diracx-logic` should embed Dirac specificities. It should encapsulate the logic of the services and should call `diracx-db` to interact with databases.
 - `diracx-db` should contain atomic methods (complex logic is expected to be located in `diracx-db`).
+
+### Error handling boundaries
+
+- `diracx-db` and `diracx-logic` should raise domain exceptions (typically `DiracError` subclasses), and should not depend on `fastapi`.
+- `diracx-routers` is the HTTP boundary and should translate known backend/domain exceptions into `HTTPException` with the desired status code, message, and headers.
+- `DiracError` handling registered at app level (see [diracx-routers/src/diracx/routers/factory.py](../../../diracx-routers/src/diracx/routers/factory.py), especially `app.add_exception_handler(DiracError, cast(handler_signature, dirac_error_handler))`) is a fallback safety net for uncaught domain exceptions, not a replacement for explicit mapping when the router needs specific HTTP behavior.
+- Protocol-specific responses can be handled directly in routers when `HTTPException` shape is not suitable (for example OAuth2-style payloads). In that case `DiracHttpResponseError` shall be used.
```
aldbr: If you remove `DiracHttpResponseError` don't forget to adjust the documentation here

diracx-core/src/diracx/core/exceptions.py
```diff
@@ -19,13 +18,10 @@
...
-        self.status_code = status_code
-        self.data = data
-
-
 class DiracError(RuntimeError):
+    # Fallback HTTP representation used by the app-level DiracError handler.
+    # Routers are expected to catch DiracError and raise HTTPException
+    # explicitly when endpoint-specific status codes/messages/headers are needed.
     http_status_code = HTTPStatus.BAD_REQUEST  # 400
     http_headers: dict[str, str] | None = None
```
aldbr: I think ideally `DiracError` should not contain anything related to HTTP as it's supposed to be isolated from the routers (`diracx-logic` could potentially be used with another layer not talking HTTP).
Here I would just propagate the `detail`.

Any opinion?

aldbr (dismissed):
Thanks a lot for your PR!
The documentation looks reasonable. I just a few comments.

Note: the changes you made, or my suggestions, are breaking changes, because extensions could, in theory, already rely on these `DiracHttpResponseError` and `DiracError`.

In practice, only LHCb is running an extension in production as of now so it should be ok. We would just need to double check on our side + ask other VOs running diracx extensions in certification if they are already relying on the exceptions.

mmascher:
@aldbr I didn't realize moving `DiracHttpResponseError` could be a breaking change! I actually wanted to discuss this on Friday afternoon over Zoom, but as you know, my washing machine repair guy got in the way. 😄

Leaving it in `exceptions.py` and adding a comment saying it should only be used in routers (and only when `HTTPException` isn't sufficient) was the other option I considered. I'm happy to go back to that if you're OK with it.

Let's make sure my very first contribution is safe and doesn't break extensions or upset people. 😄

diracx-core/src/diracx/core/exceptions.py
mmascher: The worry is that one of the modules in the business logic throws a DiracError (as it should) that is not catched explicitly in the router through a:

```
try:
...
except SomeDiracError:
   raise HTTPException()
```

In that case the default exception handler added here:

https://github.com/DIRACGrid/diracx/blob/main/diracx-routers/src/diracx/routers/factory.py#L329

Might not have all the necessary information when http_response_handler is called. that can be easily fixed by directly using HTTPStatus.BAD_REQUEST here:

https://github.com/DIRACGrid/diracx/blob/main/diracx-routers/src/diracx/routers/factory.py#L431

But then again if any extension exception class currently defines http_status_code or http_headers, a hard removal could be problematic(?). But I could use some fallback lookup for that:

```
status_code = getattr(exc, "http_status_code", HTTPStatus.BAD_REQUEST)
headers = getattr(exc, "http_headers", None)
```

diracx-routers/src/diracx/routers/exceptions.py
mmascher: I couldn't find any indication in the FastAPI documentation that this pattern is recommended:

[https://fastapi.tiangolo.com/tutorial/handling-errors/](https://fastapi.tiangolo.com/tutorial/handling-errors/)

On the other hand, it seems perfectly acceptable to return a `JSONResponse` directly:

[https://fastapi.tiangolo.com/advanced/response-directly/#using-the-jsonable-encoder-in-a-response](https://fastapi.tiangolo.com/advanced/response-directly/#using-the-jsonable-encoder-in-a-response)

Looking at the history, `DiracHttpResponse` was introduced here:

[https://github.com/DIRACGrid/diracx/commit/5c0a87d892e1a4a5f3780ae3139b95e14bf22853](https://github.com/DIRACGrid/diracx/commit/5c0a87d892e1a4a5f3780ae3139b95e14bf22853)

If it's only used in this one place, I would lean towards returning a `JSONResponse` directly. Right now we're raising an exception just so the exception handler can turn it into a `JSONResponse`, which feels like an unnecessary indirection. I think we could simplify things by returning the `JSONResponse` directly.

diracx-routers/src/diracx/routers/exceptions.py
```diff
@@ -0,0 +1,9 @@
...
+    def __init__(self, status_code: int, data):
+        self.status_code = status_code
+        self.data = data
```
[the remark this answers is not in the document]
mmascher: This is not relevant anymore since I removed DiracHttpResponseError

docs/dev/reference/coding-conventions.md
mmascher: Done here: https://github.com/DIRACGrid/diracx/pull/988/changes/4fe8c1036626b645c6abea63f2178e3c0c95d189#diff-e8bb4ddf80ae9a2329f22d2838463a8f29febb44d0f4e5eb076c3207e88f19c0L144

diracx-core/src/diracx/core/exceptions.py
mmascher: done in https://github.com/DIRACGrid/diracx/pull/988/changes/279e32e3a8e2b5cf64836f304bdc82ddef83e340

diracx-routers/src/diracx/routers/exceptions.py
mmascher: Done in https://github.com/DIRACGrid/diracx/pull/988/changes/4fe8c1036626b645c6abea63f2178e3c0c95d189

mmascher:
@aldbr I think I am done here, I added commits as a reference and then solved the various threads.

mmascher:
FYI, the test is failing because of disk space: https://github.com/DIRACGrid/diracx/actions/runs/29428374798/job/87396639449?pr=988#step:8:223

fstagni (approved):

fstagni:
Even as a "breaking" change, this could really only affect lhcbdiracx. I go ahead and merge then.

PR #990, opened 2026-07-14
fix(api): add bulk_upsert method and unit tests for opensearch BaseOSDB class

### LINKED ISSUE #401: Support upserts to multiple documents for opensearch
This feature would be helpful for upserting job parameters for multiple jobs in bulk. At the moment I have to do:

https://github.com/DIRACGrid/diracx/blob/7c260ccbcac8224d1fe5699eca38b5128ea5fa15/diracx-routers/src/diracx/routers/jobs/status.py#L174-L180

See: https://github.com/DIRACGrid/diracx/pull/397#discussion_r1950658972_

vkuznet:
This PR fixes issue #401 by introducing the `bulk_upsert` API to BaseOSDB class. It is also complement by full set of unit tests for BaseOSDB class which were missing. The unit test introduces mock client and different classes for different use cases.

Depends on https://github.com/DIRACGrid/diracx/pull/1007/ 

Please note: it is my first PR and I'm happy to adjust it according to requirements/guidelines of DiracX community.


diracx-db/src/diracx/db/os/utils.py
```diff
@@ -196,6 +197,35 @@ async def upsert(self, vo: str, doc_id: int, document: Any) -> None:
...
+                "retry_on_conflict": 10,
+            }
+            for vo, doc_id, document in documents
+        )
+
+        success, errors = await async_bulk(
+            self.client,
+            actions,
+            raise_on_error=False,
+            raise_on_exception=False,
```
fstagni, line 221: What is the rationale for this?

fstagni (commented):
In the original issue (#401) it was mentioned a specific use case. Can you already verify if this fix covers that? You can add it directly in this PR.

diracx-db/src/diracx/db/os/utils.py
vkuznet: According to [aysnc_bulk documentation](https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/_async/helpers/actions.py#L267)

> ... Note that by default we raise a ``BulkIndexError`` when we encounter an error so options like ``stats_only`` only+ apply when ``raise_on_error`` is set to  ``False``.

So, it is question of policy regarding errors which I frankly don't know.

vkuznet:
> In the original issue (#401) it was mentioned a specific use case. Can you already verify if this fix covers that? You can add it directly in this PR.

@fstagni regarding issue #401. I checked the main repository code and what is described in this issue is not there the issue is based on commit outside of any branch, most likely in fork outside of the repository https://github.com/DIRACGrid/diracx/blob/7c260ccbcac8224d1fe5699eca38b5128ea5fa15/diracx-routers/src/diracx/routers/jobs/status.py#L174-L180). Therefore, before adding anything to this PR with usage of bulk insert I need to know if it is in fact required by code from main repository, see [utils.py from main branch](https://github.com/DIRACGrid/diracx/blob/main/diracx-routers/src/diracx/routers/jobs/status.py)

fstagni:
Hi, since #410 was created, the mentioned code has been moved and refactored, and can now be found in https://github.com/DIRACGrid/diracx/blob/main/diracx-logic/src/diracx/logic/jobs/status.py#L643. The issue anyway still holds, can you have a look?

diracx-db/src/diracx/db/os/utils.py
fstagni: In https://github.com/DIRACGrid/diracx/pull/1007/changes#r3734978424 @aldbr added a `DocumentUpsertError` exception which I believe could be used here too

vkuznet: @fstagni since this PR is not yet merged and I can't yet use in my PR. Do you want me to wait until it is merged or use a different exception?

fstagni: I think it's better to wait until that one is merged.

vkuznet: For the record, I'm awaiting https://github.com/DIRACGrid/diracx/pull/1007/ to be merged and then will provide code depending on `DocumentUpsertError` exception.

fstagni: Just merged #1007

vkuznet: @fstagni , I added necessary code. Turns out that further refactoring was required to satisfy tests needs. Now tests are completed and you can proceed with review.

diracx-db/src/diracx/db/os/utils.py
```diff
@@ -213,6 +214,35 @@ async def upsert(self, vo: str, doc_id: int, document: Any) -> None:
             response,
         )
 
+    async def bulk_upsert(
```
aldbr, line 217: `bulk_upsert` would need to overridden within `job_parameters_db` because we are inserting a `JobID` and a `timestamp`: https://github.com/DIRACGrid/diracx/blob/202f84ce1b35b882f03a08530b624200b449643f/diracx-db/src/diracx/db/os/job_parameters.py#L37-L43

Here it would not work I think (and it looks like it's not spotted within the tests).

I actually wonder whether `upsert` is useful now that we have `bulk_upsert`. 
I would suggest we just drop `upsert` and replace it everywhere with `bulk_upsert`, what do you think?

diracx-logic/src/diracx/logic/jobs/status.py
aldbr, line 225: `bulk_upsert` could (should) be used here too I think

diracx-db/src/diracx/db/os/utils.py
```diff
@@ -213,6 +214,35 @@ async def upsert(self, vo: str, doc_id: int, document: Any) -> None:
...
+                "retry_on_conflict": 10,
+            }
+            for vo, doc_id, document in documents
+        )
+
+        success, errors = await async_bulk(
+            self.client,
+            actions,
+            raise_on_error=False,
+            raise_on_exception=False,
```
aldbr, line 238: I'm just wondering what happens if there is a connection issue with the DB and no exception is raised.
I guess you would get 0 success, N errors but would get any information to know that there is an issue with the DB itself?

aldbr: Also I am wondering whether it would make sense to use `max_retries` and `initial_backoff`: 
https://github.com/opensearch-project/opensearch-py/blob/213b7d6b2890c19bc83ebce0a9886d7571760240/opensearchpy/_async/helpers/actions.py#L188-L192

diracx-logic/src/diracx/logic/jobs/status.py
```diff
@@ -674,15 +674,16 @@ async def _insert_parameters(
...
+    documents = (
+        (job_id_to_vo[job_id], job_id, job_params)
+        for job_id, job_params in updates.items()
+    )
+
+    _, errors = await job_parameters_db.bulk_upsert(documents)
+    if errors:
+        for error in errors:
+            logger.error("bulk insert error %s", error)
+        raise DocumentUpsertError("Failed to perform bulk insert operation")
```
aldbr, line 686: I assume this piece of code is generic because will be reused every time there is an error.
Wouldn't it make sense to raise the `DocumentUpsertError` from `diracx-db` itself? So that this part of the code only lives in `db/os/utils` and is automatically reused by the callers

diracx-db/tests/utils/test_utils.py
```diff
@@ -0,0 +1,509 @@
...
+    fields = {
+        "job_id": {"type": "long"},
+        "status": {"type": "keyword"},
+        "timestamp": {"type": "date"},
+        "vo": {"type": "keyword"},
+    }
+    index_prefix = "dummy"
+
+    def index_name(self, vo: str, doc_id: int) -> str:
+        return f"{self.index_prefix}-{vo}-{doc_id % 10}"
```
aldbr: Any reason for not reusing `diracx-testing` `DummyOSDB`? https://github.com/DIRACGrid/diracx/blob/d72016500fb6cfd73350aa7b2582c5dd2cdcf2d2/diracx-testing/src/diracx/testing/dummy_osdb.py

Like in https://github.com/DIRACGrid/diracx/blob/d72016500fb6cfd73350aa7b2582c5dd2cdcf2d2/diracx-db/tests/opensearch/test_search.py#L22

diracx-db/tests/utils/test_utils.py
aldbr, line 1: Isn't there some duplication with https://github.com/DIRACGrid/diracx/blob/d72016500fb6cfd73350aa7b2582c5dd2cdcf2d2/diracx-db/tests/opensearch/test_search.py?

diracx-logic/tests/conftest.py
aldbr, line 1: I think you can remove this `conftest.py` file and add your `mock_osdb` fixture within `os_db.py`, which is already including in `pyproject.toml` of the `diracx-*` packages. See in https://github.com/DIRACGrid/diracx/blob/d72016500fb6cfd73350aa7b2582c5dd2cdcf2d2/diracx-logic/pyproject.toml#L59

Though it looks like this is not documented...

diracx-db/tests/conftest.py
aldbr, line 1: Same comment here

aldbr (changes requested):
Thanks @vkuznet!
Just a few additional comments, sorry for the delay we take to review PRs...

diracx-db/src/diracx/db/os/utils.py
vkuznet: This is fixed now in https://github.com/DIRACGrid/diracx/pull/990/commits/6d2984c1d6ea0b8062b809feaa0cccdb82bc14f5

But I against dropping `upsert` in favor of `bulk_upsert` for simple reason. The code, e.g. https://github.com/DIRACGrid/diracx/blob/main/diracx-logic/src/diracx/logic/jobs/status.py#L225, uses external for loop and insert each document individually while going through that loop. To use `bulk_upsert` would require in this place to either use generator or allocate more memory to collect all documents and then insert them in bulk. There are cases when one API is preferable vs another. Since I can't find usage of generators I think the `upsert` has its place in a code.

diracx-logic/src/diracx/logic/jobs/status.py
vkuznet: It can but it would either refactor code to use generators or use additional memory allocation to collect all documents. Either task seems beyond this PR scope.

diracx-db/src/diracx/db/os/utils.py
vkuznet: in order to use this features it would be desired to have them configurable rather using hard-coded defaults. At the moment (based on my limited scope of the code) I don't know how configuration work and if desired this can be done through a separate issue/PR.

diracx-db/tests/utils/test_utils.py
vkuznet: I was not aware of another DummyOSDB implementation. Now it is fixed by using existing one in this commit: https://github.com/DIRACGrid/diracx/pull/990/commits/c2b68aa2735b82cdfd85b7ce56850fae91a026df

diracx-logic/src/diracx/logic/jobs/status.py
vkuznet: this is architectural choice, your suggestion to through exception in base bulk_upsert API while mine is to pass errors to upstream and let this code decide what to do. If you can provide specific use-case when exception is better I'll be happy to move this part down the stream and through exception in a base class.

diracx-logic/tests/conftest.py
vkuznet: ok, done in https://github.com/DIRACGrid/diracx/pull/990/commits/188948da771e5d1a6503e3c5997fe457f5e28ab6

diracx-db/tests/conftest.py
vkuznet: done in https://github.com/DIRACGrid/diracx/pull/990/commits/188948da771e5d1a6503e3c5997fe457f5e28ab6

diracx-db/src/diracx/db/os/utils.py
vkuznet: > I'm just wondering what happens if there is a connection issue with the DB and no exception is raised. I guess you would get 0 success, N errors but would get any information to know that there is an issue with the DB itself?

this is up to upstream code. Since it is external we can't reliably tell what is current and future functionality would be.

vkuznet:
@aldbr , I hope I replied and addressed all your concern. But I want to make clear about my last commit: https://github.com/DIRACGrid/diracx/pull/990/commits/eacac9acba4dc307d152b6f8c3046f5f7c4ce6e3

It was required to fix failing integration test `diracx-db/tests/utils/test_utils.py`. Moreover, in my working shell environment this commit was not required but on github CI/CD this particular test file was failing without this commit. Anyway, the commit inject `mock_client` which can be done in different ways, originally I did it through `conftest.py` file but later you suggested to use `pyproject.toml` addition. Since I don't know and control github CI/CD env I can't say for sure  why failure occurs using `pyproject.toml` methon. But once I put back `mock_client` via `conftest.py` the test was fine.

Feel free to proceed review and make further step.

diracx-db/src/diracx/db/os/job_parameters.py
```diff
@@ -41,3 +42,15 @@ def upsert(self, vo, doc_id, document):
...
+    async def bulk_upsert(
+        self,
+        documents: Iterable[tuple[str, int, dict[str, Any]]],
+    ) -> tuple[int, list[Any]]:
+        """bulk_upsert API implementation."""
+        transformed = []
+        for vo, doc_id, document in documents:
+            timestamp = int(datetime.now(tz=UTC).timestamp() * 1000)
+            document = {"JobID": doc_id, "timestamp": timestamp, **document}
+            transformed.append((vo, doc_id, document))
```
aldbr: Minor: do you need to build 1 timestamp per document?
What about?

```suggestion
        transformed = []
        timestamp = int(datetime.now(tz=UTC).timestamp() * 1000)
        for vo, doc_id, document in documents:
            document = {"JobID": doc_id, "timestamp": timestamp, **document}
            transformed.append((vo, doc_id, document))
```

Also is there any reason for having the `document` intermediate variable here?

diracx-db/src/diracx/db/os/utils.py
aldbr: > in order to use this features it would be desired to have them configurable rather using hard-coded defaults. At the moment (based on my limited scope of the code) I don't know how configuration work and if desired this can be done through a separate issue/PR.

We can indeed wait to see how it behaves and see if we get any other use case

diracx-db/tests/utils/test_utils.py
aldbr: Have you seen this comment?

diracx-logic/src/diracx/logic/jobs/status.py
aldbr: I think it's in the PR scope, as it's the same pattern described in #401 

In terms of memory, I assume it would be around the same size as `job_attribute_updates` and it would only live for the duration of the call. I am not sure I understand the issue, please feel free to contradict if you think I am missing something.

And IIUC, it would be much faster to be processed rather than having single calls to `OpenSearch` as of today.

aldbr (changes requested):

diracx-logic/src/diracx/logic/jobs/status.py
aldbr: Fine, let's see once we have other use cases

aldbr:
> @aldbr , I hope I replied and addressed all your concern. But I want to make clear about my last commit: [eacac9a](https://github.com/DIRACGrid/diracx/commit/eacac9acba4dc307d152b6f8c3046f5f7c4ce6e3)
> 
> It was required to fix failing integration test `diracx-db/tests/utils/test_utils.py`. Moreover, in my working shell environment this commit was not required but on github CI/CD this particular test file was failing without this commit. Anyway, the commit inject `mock_client` which can be done in different ways, originally I did it through `conftest.py` file but later you suggested to use `pyproject.toml` addition. Since I don't know and control github CI/CD env I can't say for sure why failure occurs using `pyproject.toml` methon. But once I put back `mock_client` via `conftest.py` the test was fine.
> 
> Feel free to proceed review and make further step.

Did you also add it to the main `pyproject.toml`?

But what about moving `mock_client` from `testing.mock_osdb.py` to `testing.osdb.py`? You wouldn't have to have any `pytest` import in `mock_osdb.py` and normally everything should work as expected because `osdb` is already included in the various `pyproject.toml`.

Thanks a lot for your work and sorry for the delay...

vkuznet:
> > @aldbr , I hope I replied and addressed all your concern. But I want to make clear about my last commit: [eacac9a](https://github.com/DIRACGrid/diracx/commit/eacac9acba4dc307d152b6f8c3046f5f7c4ce6e3)
> > It was required to fix failing integration test `diracx-db/tests/utils/test_utils.py`. Moreover, in my working shell environment this commit was not required but on github CI/CD this particular test file was failing without this commit. Anyway, the commit inject `mock_client` which can be done in different ways, originally I did it through `conftest.py` file but later you suggested to use `pyproject.toml` addition. Since I don't know and control github CI/CD env I can't say for sure why failure occurs using `pyproject.toml` methon. But once I put back `mock_client` via `conftest.py` the test was fine.
> > Feel free to proceed review and make further step.
> 
> Did you also add it to the main `pyproject.toml`?
> 
> But what about moving `mock_client` from `testing.mock_osdb.py` to `testing.osdb.py`? You wouldn't have to have any `pytest` import in `mock_osdb.py` and normally everything should work as expected because `osdb` is already included in the various `pyproject.toml`.
> 
> Thanks a lot for your work and sorry for the delay...

ok, I'll try again and ensure that that it is added to `pyproject.toml`.

diracx-db/tests/utils/test_utils.py
vkuznet: Ahh, sorry, I forgot to reply here. The tests are complementary but not the same. The test_utils.py verifies what the code sends and how it is wired, while test_search.py verifies what users get back, across different backend. The former is pure unit tests while latter seems more like integration test.

vkuznet:
@aldbr , I applied changes you requested, and I'm glad that your hint about `pyproject.toml` fixed issue with mocking module. Now I see that all CI/CD tests pass through and this PR is ready again for hopefully final review.

PR #993, opened 2026-07-16
docs: modify rest style comments with doc convert to follow google style.

### LINKED ISSUE #927: Enhancement: Harmonize Docstring convention to Google Style (Native MkDocs Support & CI Enforcement)
## Background and Current State

Currently, the DiracX codebase suffers from an inconsistency in its docstring conventions. A review of the code reveals a mix of **Google Style** docstrings and legacy **reStructuredText (reST)** style docstrings. 

As DiracX utilizes **MkDocs** (and Markdown) for its documentation rendering rather than Sphinx, this mix creates both readability issues for developers and architectural friction for our documentation generators (like `mkdocstrings`).

## The Problem with the Current Inconsistency

1. **Native Compatibility:** `mkdocstrings` is fundamentally designed to parse the Google docstring convention natively and seamlessly convert it into Markdown. Conversely, parsing reST tags (`:param:`, `:type:`) in an MkDocs environment is an architectural mismatch that requires heavy workarounds and often results in poorly formatted HTML output.
2. **Readability:** Having two different styles across the repository increases cognitive load for developers reading the raw code.
3. **DRY Principle Violation:** reST forces developers to duplicate type hints inside the docstring, ignoring the modern Python type annotations (PEP 484) that DiracX heavily uses.

## Pros & Cons Comparison

To align the team on the best path forward, here is a breakdown of why standardizing purely on the Google convention is the optimal architectural choice for DiracX:

| Feature / Aspect | Google Style (Proposed Standard) | reStructuredText (Current Legacy) |
| :--- | :--- | :--- |
| **MkDocs / Markdown Integration** | **Native & Flawless** (via `mkdocstrings`) | **Poor** (Requires complex plugins/hacks) |
| **Raw Code Readability** | **High** (Clean sections: `Args:`, `Returns:`) | **Low** (Cluttered with explicit `:tags:`) |
| **DRY (Type Hinting)** | **Excellent** (Extracts types from signatures) | **Redundant** (Types must be written twice) |
| **Contributor Onboarding** | **Intuitive** (Lower barrier to entry) | **Steep curve** (Syntax is strict and verbose) |
| **Legacy Portability** | N/A | **Pro:** Easier copy/pasting from legacy DIRAC. *(Note: As DiracX is a modern rewrite, this benefit is practically negligible).* |

## Proposed Solution & Tooling

We should officially deprecate the reST docstring style and mandate the **Google Style** across the entire DiracX repository.

To enforce this without relying on manual code reviews, we can leverage our existing linter. Since the project already relies on **Ruff**, we can enforce the Google convention automatically by adding the following to our `pyproject.toml`:

```toml
[tool.ruff.lint.pydocstyle]
convention = "google"
```

This will integrate the docstring format check directly into our CI/CD pipeline, ensuring that all future Pull Requests maintain perfect consistency.

## Migration Strategy

* Change all non-Google style docstrings to **Google style**.
* Apply the Ruff configuration to enforce the rule for all new code.

## Feedback Requested

I would like to bring this up during a DDev to agree on standardizing our documentation architecture. Feedback on the automated enforcement via Ruff is highly appreciated!

martynia:
A minimal change, only convert rest style comments to google style comments. Don't modify comments themselves, when possible. Fixes #927

### DIFF
--- a/diracx-core/src/diracx/core/sources.py
+++ b/diracx-core/src/diracx/core/sources.py
@@ -80,22 +80,14 @@ class CacheableSource(Generic[T], metaclass=ABCMeta):
         """
 
     def read(self) -> T:
-        """Load the source from the backend with appropriate caching.
-
-        :raises: diracx.core.exceptions.NotReadyError if the source is being loaded still
-        :raises: git.exc.BadName if version does not exist
-        """
+        """Load the source from the backend with appropriate caching."""
         hexsha = self._revision_cache.get(
             "latest_revision", self._read_work, blocking=True
         )
         return self._content_cache[hexsha]
 
     async def read_non_blocking(self) -> T:
-        """Load the source from the backend with appropriate caching.
-
-        :raises: diracx.core.exceptions.NotReadyError if the source is being loaded still
-        :raises: git.exc.BadName if version does not exist
-        """
+        """Load the source from the backend with appropriate caching."""
         hexsha = self._revision_cache.get(
             "latest_revision", self._read_work, blocking=False
         )
--- a/diracx-db/src/diracx/db/sql/auth/db.py
+++ b/diracx-db/src/diracx/db/sql/auth/db.py
@@ -160,7 +160,7 @@ class AuthDB(BaseSQLDB):
         Returns the scope field for the given user_code
 
         Raises:
-            NoResultFound if no such user code currently Pending
+            NoResultFound: if no such user code currently Pending
         """
         stmt = select(DeviceFlows.scope).where(
             DeviceFlows.user_code == user_code,
@@ -271,9 +271,10 @@ class AuthDB(BaseSQLDB):
     async def authorization_flow_insert_id_token(
         self, uuid: str, id_token: dict[str, str], max_validity: int
     ) -> tuple[str, str]:
-        """Return code, redirect_uri.
+        """Return the authorization code and redirect URI.
 
-        Raises: AuthorizationError if no such uuid or status not pending.
+        Raises:
+            AuthorizationError: If no such UUID exists or the flow is not pending.
         """
         # Hash the code to avoid leaking information
         code = secrets.token_urlsafe()
--- a/diracx-db/src/diracx/db/sql/auth/db.py
+++ b/diracx-db/src/diracx/db/sql/auth/db.py
@@ -159,7 +159,7 @@ class AuthDB(BaseSQLDB):
 
         Returns the scope field for the given user_code
 
-        :raises:
+        Raises:
             NoResultFound if no such user code currently Pending
         """
         stmt = select(DeviceFlows.scope).where(
@@ -273,7 +273,7 @@ class AuthDB(BaseSQLDB):
     ) -> tuple[str, str]:
         """Return code, redirect_uri.
 
-        :raises: AuthorizationError if no such uuid or status not pending.
+        Raises: AuthorizationError if no such uuid or status not pending.
         """
         # Hash the code to avoid leaking information
         code = secrets.token_urlsafe()
--- a/diracx-db/src/diracx/db/sql/job/db.py
+++ b/diracx-db/src/diracx/db/sql/job/db.py
@@ -205,7 +205,7 @@ class JobDB(BaseSQLDB):
             update_timestamp: if True, update the LastUpdate to now
 
         Returns:
-          rowcount
+            rowcount
 
         """
         # Check that all we always update the same set of properties
--- a/diracx-db/src/diracx/db/sql/job/db.py
+++ b/diracx-db/src/diracx/db/sql/job/db.py
@@ -200,10 +200,12 @@ class JobDB(BaseSQLDB):
 
         All the jobs must update the same properties.
 
-        :param properties: {job_id : {prop1: val1, prop2:val2}
-        :param update_timestamp: if True, update the LastUpdate to now
+        Args:
+            properties: {job_id : {prop1: val1, prop2:val2}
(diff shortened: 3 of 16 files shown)


diracx-db/src/diracx/db/sql/auth/db.py
```diff
@@ -159,7 +159,7 @@ async def device_flow_validate_user_code(
 
         Returns the scope field for the given user_code
 
-        :raises:
+        Raises:
             NoResultFound if no such user code currently Pending
```
devlink42:
```suggestion
            NoResultFound: if no such user code currently Pending
```

diracx-db/src/diracx/db/sql/auth/db.py
```diff
@@ -273,7 +273,7 @@ async def authorization_flow_insert_id_token(
     ) -> tuple[str, str]:
         """Return code, redirect_uri.
 
-        :raises: AuthorizationError if no such uuid or status not pending.
+        Raises: AuthorizationError if no such uuid or status not pending.
```
devlink42:
```suggestion
        Raises:
            AuthorizationError: if no such uuid or status not pending.
```

diracx-db/src/diracx/db/sql/job/db.py
```diff
@@ -200,10 +200,12 @@ async def set_properties(
...
 
-        :param properties: {job_id : {prop1: val1, prop2:val2}
-        :param update_timestamp: if True, update the LastUpdate to now
+        Args:
+            properties: {job_id : {prop1: val1, prop2:val2}
+            update_timestamp: if True, update the LastUpdate to now
 
-        :return rowcount
+        Returns:
+          rowcount
```
devlink42, line 209:
```suggestion
        Returns:
            rowcount
```

diracx-logic/src/diracx/logic/jobs/status.py
```diff
@@ -103,7 +103,8 @@ async def set_job_statuses(
     logging information in the JobLoggingDB. The status dict has datetime
     as a key and status information dictionary as values.
 
-    :raises: JobNotFound if the job is not found in one of the DBs
+    Raises:
+      JobNotFound if the job is not found in one of the DBs
```
devlink42:
```suggestion
        JobNotFound: if the job is not found in one of the DBs
```

diracx-logic/src/diracx/logic/auth/utils.py
```diff
@@ -243,7 +243,7 @@ def parse_and_validate_scope(
         * properties are known
     return dict with group and properties.
 
-    :raises:
+    Raises:
         * ValueError in case the scope isn't valid
```
devlink42:
```suggestion
        ValueError: in case the scope isn't valid
```

diracx-core/src/diracx/core/sources.py
```diff
@@ -80,22 +80,14 @@ def read_raw(self, hexsha: str, modified: datetime) -> T:
         """
 
     def read(self) -> T:
-        """Load the source from the backend with appropriate caching.
-
-        :raises: diracx.core.exceptions.NotReadyError if the source is being loaded still
-        :raises: git.exc.BadName if version does not exist
```
aldbr, line 86: `Raises` have been deleted, is that expected?

diracx-core/src/diracx/core/sources.py
```diff
@@ -80,22 +80,14 @@ def read_raw(self, hexsha: str, modified: datetime) -> T:
...
         hexsha = self._revision_cache.get(
             "latest_revision", self._read_work, blocking=True
         )
         return self._content_cache[hexsha]
 
     async def read_non_blocking(self) -> T:
-        """Load the source from the backend with appropriate caching.
-
-        :raises: diracx.core.exceptions.NotReadyError if the source is being loaded still
-        :raises: git.exc.BadName if version does not exist
```
aldbr, line 97: Same here, they have been deleted

aldbr (dismissed):

martynia:
The converter is not perfect. I spotted a few deletions, I missed those
apparently.

diracx-core/src/diracx/core/sources.py
devlink42: I don't think so because there is no explicit 'raise' in the core of the function, so we don't normally have to put it in, do we?

diracx-logic/src/diracx/logic/auth/utils.py
martynia: Added a colon after ValueError:

diracx-logic/src/diracx/logic/jobs/status.py
martynia: Indent 4 spaces and add a colon after JobNotFound:

diracx-core/src/diracx/core/sources.py
martynia: I think @devlink42 is right here.

diracx-core/src/diracx/core/sources.py
martynia: No explicit rase.

diracx-logic/src/diracx/logic/auth/utils.py
devlink42: Yes, also exactly!

devlink42 (approved):

diracx-db/src/diracx/db/sql/job/db.py
martynia: Is it still pending? I fixed it I believe.

fstagni (approved):

PR #995, opened 2026-07-16
feat(tasks): add dummy job executor

### LINKED ISSUE #964: [Feature]: Dummy Job Executor
### User Story

As a developer, I want to be able to use DiracX to submit and execute a very simple job without having to through DIRAC, so that I can use it as a standalone for testing purposes

### Feature Description

I want a dummy job executor as a diracx task.

### Definition of Done

- [x] task that prints "I am executing \<jobid\>"
- [x] unit tests
- [x] documentation on how to use it

### Alternatives Considered

_No response_

### Related Issues

_No response_

### Additional Context

_No response_

anpicci:
## Summary

- add `DummyJobExecutorMonitorTask`, a settings-controlled periodic task that:
  - is disabled by default;
  - when enabled, searches for jobs in `Received`, moves them to `Waiting`, and schedules one executor task per job;
- add `DummyJobExecutorTask`, a one-shot task that simulates job execution through the valid state sequence `Waiting → Matched → Running → Done`, while holding a per-job `MutexLock`;
- register and publicly export both tasks as `jobs:DummyJobExecutorMonitorTask` and `jobs:DummyJobExecutorTask`;
- add administrator documentation for the dummy-executor environment settings;
- enable the monitor explicitly at a 10-second interval in `run_local.sh`;
- provide the corresponding demo-deployment enablement in DIRACGrid/diracx-charts#286;
- add focused tests for scheduling activation, interval validation, locking, status transitions, task fan-out, and the no-job case.

## Motivation

This provides a standalone dummy job-execution pipeline for exercising the DiracX task framework end to end—job discovery, status transitions, task fan-out, and per-job locking—without requiring execution through legacy DIRAC. The monitor remains opt-in so production/default deployments do not schedule it automatically.

Closes #964.

## Implementation notes

- `DummyJobExecutorMonitorTask` uses `Size.SMALL` and does not define a retry policy, since an enabled monitor runs periodically.
- `DummyJobExecutorTask` uses `Size.LARGE` and `ExponentialBackoff(base_delay_seconds=10, max_retries=3)`.
- The executor applies the `Matched → Running → Done` transitions in one `set_job_statuses` call with increasing timestamps, because a direct transition to `Done` would be rejected by the state machine.
- All database dependencies (`JobDB`, `JobLoggingDB`, `TaskQueueDB`, and `JobParametersDB`) and `Config` are injected through the task framework.
- The monitor is controlled through:
  - `DIRACX_TASKS_DUMMY_JOB_EXECUTOR_ENABLED` (default: `false`);
  - `DIRACX_TASKS_DUMMY_JOB_EXECUTOR_INTERVAL_SECONDS` (default: `10`, required to be positive).
- `run_local.sh` explicitly enables the monitor at a 10-second interval. The standard demo deployment is handled by the companion charts PR, DIRACGrid/diracx-charts#286.
- Deliberately out of scope: JDL handling, real matching or pilots, and real job execution.

## Validation

- focused dummy-executor tests: 6 passed;
- full `diracx-tasks` suite: 71 passed;
- all Pixi-dependent pre-commit hooks passed with `--all-files`;
- DIRACX and Gubbins client-generation checks passed with no retained generated-client changes;
- Basic Tests, Integration Tests, Pixi Lock, and Deployment GitHub Actions workflows all pass.

### DIFF
--- a/diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
+++ b/diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
@@ -1,24 +1,145 @@
-"""Task that simulates executing a job."""
+"""Tasks that simulate the execution of jobs, for use in demo deployments."""
 
 from __future__ import annotations
 
 import dataclasses
 import logging
+from datetime import UTC, datetime, timedelta
 
-from diracx.tasks.plumbing.base_task import BaseTask
-from diracx.tasks.plumbing.enums import Size
+from diracx.core.models import (
+    JobStatus,
+    JobStatusUpdate,
+    ScalarSearchOperator,
+    ScalarSearchSpec,
+)
+from diracx.db.os import JobParametersDB
+from diracx.db.sql import JobDB, JobLoggingDB, TaskQueueDB
+from diracx.logic.jobs import set_job_statuses
+from diracx.tasks.plumbing.base_task import BaseTask, PeriodicBaseTask
+from diracx.tasks.plumbing.depends import Config
+from diracx.tasks.plumbing.enums import Priority, Size
+from diracx.tasks.plumbing.lock_registry import JOB
+from diracx.tasks.plumbing.locks import BaseLock, MutexLock
+from diracx.tasks.plumbing.retry_policies import ExponentialBackoff
+from diracx.tasks.plumbing.schedules import IntervalSeconds
 
 logger = logging.getLogger(__name__)
+
+MINOR_STATUS = "DummyExecutor"
+
+
+@dataclasses.dataclass
+class DummyJobExecutorMonitorTask(PeriodicBaseTask):
+    """Periodically pick up newly received jobs and hand them to the dummy executor.
+
+    Every job in ``Received`` state is moved to ``Waiting`` and a one-shot
+    ``DummyJobExecutorTask`` is scheduled for it.
+    """
+
+    priority = Priority.BACKGROUND
+    size = Size.MEDIUM
+    retry_policy = ExponentialBackoff(base_delay_seconds=300, max_retries=3)
+    default_schedule = IntervalSeconds(10)
+
+    async def execute(
+        self,
+        config: Config,
+        job_db: JobDB,
+        job_logging_db: JobLoggingDB,
+        task_queue_db: TaskQueueDB,
+        job_parameters_db: JobParametersDB,
+    ) -> int:
+        _, jobs = await job_db.search(
+            ["JobID"],
+            [
+                ScalarSearchSpec(
+                    parameter="Status",
+                    operator=ScalarSearchOperator.EQUAL,
+                    value=JobStatus.RECEIVED,
+                )
+            ],
+            [],
+        )
+        if not jobs:
+            return 0
+
+        job_ids = [job["JobID"] for job in jobs]
+        logger.info("Moving %d received job(s) to Waiting: %s", len(job_ids), job_ids)
+        await set_job_statuses(
+            {
+                job_id: {
+                    datetime.now(UTC): JobStatusUpdate(
+                        Status=JobStatus.WAITING,
+                        MinorStatus=MINOR_STATUS,
+                    )
+                }
+                for job_id in job_ids
+            },
+            config=config,
+            job_db=job_db,
+            job_logging_db=job_logging_db,
+            task_queue_db=task_queue_db,
+            job_parameters_db=job_parameters_db,
+        )
+
+        for job_id in job_ids:
+            await DummyJobExecutorTask(job_id=job_id).schedule()
+
+        return len(job_ids)
 
 
 @dataclasses.dataclass
 class DummyJobExecutorTask(BaseTask):
-    """Log execution of a single job."""
+    """Simulate the execution of a single job.
+
+    The job is walked through the state machine's valid path
+    ``Waiting -> Matched -> Running -> Done``: a direct jump to ``Done`` would be
+    silently rejected. ``set_job_statuses`` applies the timestamped updates in
+    chronological order, so a single call with increasing timestamps walks the
+    whole chain.
+    """
+
+    priority = Priority.NORMAL
+    size = Size.SMALL
+    retry_policy = ExponentialBackoff(base_delay_seconds=300, max_retries=3)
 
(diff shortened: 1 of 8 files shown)


diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
```diff
@@ -0,0 +1,145 @@
...
+class DummyJobExecutorMonitorTask(PeriodicBaseTask):
+    """Periodically pick up newly received jobs and hand them to the dummy executor.
+
+    Every job in ``Received`` state is moved to ``Waiting`` and a one-shot
+    ``DummyJobExecutorTask`` is scheduled for it.
+    """
+
+    priority = Priority.BACKGROUND
+    size = Size.MEDIUM
+    retry_policy = ExponentialBackoff(base_delay_seconds=300, max_retries=3)
```
chrisburr: I think we don't need a retry policy here as it's going to run every 10 seconds regardless.

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
```diff
@@ -0,0 +1,145 @@
...
+    The job is walked through the state machine's valid path
+    ``Waiting -> Matched -> Running -> Done``: a direct jump to ``Done`` would be
+    silently rejected. ``set_job_statuses`` applies the timestamped updates in
+    chronological order, so a single call with increasing timestamps walks the
+    whole chain.
+    """
+
+    priority = Priority.NORMAL
+    size = Size.SMALL
+    retry_policy = ExponentialBackoff(base_delay_seconds=300, max_retries=3)
```
chrisburr: As this is for local development 5 minutes is a very long time:

```suggestion
    retry_policy = ExponentialBackoff(base_delay_seconds=10, max_retries=3)
```

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
```diff
@@ -0,0 +1,145 @@
...
+
+    The job is walked through the state machine's valid path
+    ``Waiting -> Matched -> Running -> Done``: a direct jump to ``Done`` would be
+    silently rejected. ``set_job_statuses`` applies the timestamped updates in
+    chronological order, so a single call with increasing timestamps walks the
+    whole chain.
+    """
+
+    priority = Priority.NORMAL
+    size = Size.SMALL
```
chrisburr: Eventually this should be CPU/memory bound so I think this make more sense here: 

```suggestion
    size = Size.LARGE
```

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
```diff
@@ -0,0 +1,145 @@
...
+@dataclasses.dataclass
+class DummyJobExecutorMonitorTask(PeriodicBaseTask):
+    """Periodically pick up newly received jobs and hand them to the dummy executor.
+
+    Every job in ``Received`` state is moved to ``Waiting`` and a one-shot
+    ``DummyJobExecutorTask`` is scheduled for it.
+    """
+
+    priority = Priority.BACKGROUND
+    size = Size.MEDIUM
```
chrisburr: This should be mostly IO bound by querying the database so:

```suggestion
    size = Size.SMALL
```

diracx-tasks/tests/test_dummy_job_executor.py
```diff
@@ -0,0 +1,122 @@
+"""Tests for the dummy job executor tasks."""
```
chrisburr, line 1: I think these tests are too verbose at the moment, we should focus testing the actual logic rather than repeating the configuration.

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
```diff
@@ -0,0 +1,145 @@
...
+    retry_policy = ExponentialBackoff(base_delay_seconds=300, max_retries=3)
+    default_schedule = IntervalSeconds(10)
```
chrisburr: There should probably be no schedule by default and run_local/run_demo should override it to be every 10 seconds.

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
anpicci: Agreed. I removed the explicit retry policy from the monitor; when enabled, it now relies on its periodic schedule.

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
anpicci: Applied. The executor now uses `ExponentialBackoff(base_delay_seconds=10, max_retries=3)`.

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
anpicci: Applied. `DummyJobExecutorTask` now uses `Size.LARGE`.

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
anpicci: Applied. `DummyJobExecutorMonitorTask` now uses `Size.SMALL`.

diracx-tasks/tests/test_dummy_job_executor.py
anpicci: Agreed. I reduced the suite to six focused tests covering environment-controlled scheduling, positive interval validation, the per-job mutex, the executor state sequence, monitor fan-out, and the no-job case.

diracx-tasks/src/diracx/tasks/jobs/dummy_job_executor.py
anpicci: Implemented. The monitor is disabled by default, so the scheduler does not track it unless explicitly enabled. `run_local.sh` enables it with a 10-second interval, and the companion demo configuration is in DIRACGrid/diracx-charts#286.

ryuwd (approved):
LGTM!

Issue #1004, opened 2026-08-04
Add missing docstrings and guard against reST regressions (follow-up to #927)


aldbr:
### User Story

As a DiracX developer or extension author,
I want the public modules, classes, and functions to carry Google-style docstrings, with CI rejecting missing docstrings and legacy reST tags,
So that the mkdocstrings-generated API reference is complete and the convention adopted in #927 cannot silently regress.

### Feature Description

#927 harmonized the existing docstrings to Google style (PR #993) and enabled `[tool.ruff.lint.pydocstyle] convention = "google"`. This follow-up covers what was intentionally left out of that issue's scope:

1. Write the missing docstrings. The missing-docstring rules D100–D107 are still in the Ruff `ignore` list (`# TODO: Maybe enable these` in `pyproject.toml`). Proposal: enable **D100–D103** (missing docstring in module / class / method / function) for `src` code, and *permanently* ignore D104 (package `__init__.py`), D105 (magic methods), D107 (`__init__` — the class docstring covers construction) plus all of `tests/`, as most projects with strict docstring policies do.

2. Convert the one reST docstring that survived #993: `diracx-routers/tests/health/test_probes.py` still uses `:param:`/`:return:` tags. Ruff doesn't catch it because `convention = "google"` only disables conflicting pydocstyle checks and does not forbid reST tags.

3. Guard against reST tags coming back as it will highly likely try to come back as most of the DIRAC docstrings is using reST. Add a `pygrep` hook to `.pre-commit-config.yaml`:

   ```yaml
   - repo: local
     hooks:
       - id: forbid-rest-docstrings
         name: forbid reST docstring tags (use Google style, see #927)
         language: pygrep
         entry: '^\s*:(param|type|returns?|rtype|raises?|ivar|cvar|vartype)\b'
         types: [python]
         exclude: ^diracx-client/src/diracx/client/
   ```

Suggested migration path (following our [splitting-PRs workflow](https://diracx.diracgrid.org/en/latest/dev/explanations/splitting-prs/), to avoid a repeat of #965):

- Bootstrap PR: fix `test_probes.py`, add the pygrep hook, move the agreed D1xx ignores from the global `ignore` list into `per-file-ignores` entries (one per package, plus permanent exemptions for tests).
[...]

PR #1008, opened 2026-08-07
docs: move D100-D300 to per-file-ignores, forbid rest-docstrings. 

martynia:
A minimal, "bootstrap" PR to address #1004

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -133,3 +133,12 @@ repos:
         entry: pixi run -e default python scripts/check_init_files_precommit_hook.py --mode=forbid
         files: ^(diracx-|extensions/gubbins/gubbins-)[a-z]+/src/[a-z]+/[a-z]+/[a-z_]+/.+\.py$
         exclude: (__init__\.py$|test|_generated|__main__|/patches/)
+
+  - repo: local
+    hooks:
+      - id: forbid-rest-docstrings
+        name: forbid reST docstring tags (use Google style, see #927)
+        language: pygrep
+        entry: '^\s*:(param|type|returns?|rtype|raises?|ivar|cvar|vartype)\b'
+        types: [python]
+        exclude: ^diracx-client/src/diracx/client/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -141,4 +141,4 @@ repos:
         language: pygrep
         entry: '^\s*:(param|type|returns?|rtype|raises?|ivar|cvar|vartype)\b'
         types: [python]
-        exclude: ^diracx-client/src/diracx/client/
+        exclude: "^(diracx-client/src/diracx/client|extensions/gubbins/gubbins-client/src/gubbins/client)/"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,10 +54,6 @@ ignore = [
   "B006",
   "S101", # bandit: use of assert https://docs.astral.sh/ruff/rules/assert/
   # TODO: Maybe enable these
-  "D100",
-  "D101",
-  "D102",
-  "D103",
   "D104",
   "D105",
   "D107",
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -135,3 +133,12 @@ repos:
...
         exclude: (__init__\.py$|test|_generated|__main__|/patches/)
+
+  - repo: local
+    hooks:
+      - id: forbid-rest-docstrings
+        name: forbid reST docstring tags (use Google style, see #927)
+        language: pygrep
+        entry: '^\s*:(param|type|returns?|rtype|raises?|ivar|cvar|vartype)\b'
+        types: [python]
+        exclude: "^(diracx-client/src/diracx/client|extensions/gubbins/gubbins-client/src/gubbins/client)/"
```
aldbr, line 144: What about? 
```suggestion
        exclude: '/_generated/'
```

pyproject.toml
```diff
@@ -73,6 +69,43 @@ required-imports = ["from __future__ import annotations"]
...
+"extensions/gubbins/gubbins-core/tests/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-db/src/gubbins/db/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-db/tests/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-logic/src/gubbins/logic/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-logic/tests/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-routers/src/gubbins/routers/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-routers/tests/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-tasks/src/gubbins/tasks/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-tasks/tests/**/*.py" = ["D100", "D101", "D102", "D103"]
+"extensions/gubbins/gubbins-testing/src/gubbins/testing/**/*.py" = ["D100", "D101", "D102", "D103"]
```
aldbr, line 105: I am not fully sure this is going to work, I think these need to go in `extensions/gubbins/pyproject.toml`.
Btw, I just noticed that the following blocks should also be added there ideally:

```toml
[tool.ruff.lint.pydocstyle]
convention = "google"
```

```toml
[tool.ruff.lint]
select = [
    ...
    "D",   # pydocstyle
```

aldbr (changes requested):

aldbr:
Ideally, I would like to see at least one PR based on this one where you add the missing docstrings in a package (as you already did) and you remove the package from `[tool.ruff.lint.per-file-ignores]` just to make sure it works well before approving this one.
Can you try that please?

pyproject.toml
martynia: This _does_ work. The way I noticed a problem with gubbins was that there is a non google comment there. To silence it I added the lines above and the  error is gone.

.pre-commit-config.yaml
martynia: This suggestion does not work:
`pixi run -e pre-commit pre-commit run --all-files forbid-rest-docstrings
✨ Pixi task (pre-commit in pre-commit): pre-commit run --all-files forbid-rest-docstrings
forbid reST docstring tags (use Google style, see........................Failed
- hook id: forbid-rest-docstrings
- exit code: 1

diracx-client/src/diracx/client/patches/client/aio.py:93:        :param request: The pipeline request object to be modified.
diracx-client/src/diracx/client/patches/client/aio.py:94:        :type request: ~azure.core.pipeline.PipelineRequest
diracx-client/src/diracx/client/patches/client/aio.py:95:        :raises: :class:`~azure.core.exceptions.ServiceRequestError`
diracx-client/src/diracx/client/patches/client/sync.py:73:        :param request: The pipeline request object to be modified.
diracx-client/src/diracx/client/patches/client/sync.py:74:        :type request: ~azure.core.pipeline.PipelineRequest
diracx-client/src/diracx/client/patches/client/sync.py:75:        :raises: :class:`~azure.core.exceptions.ServiceRequestError``

fstagni:
> Ddev: un-necessary now

PR #1026, opened 2026-09-01
docs(rtd): fix broken 404-page asset paths

kjvbrt:
`site_url` pointed at an unused domain with no path, so mkdocs generated root-absolute asset links for version-less pages like 404.html. Read the Docs serves this project under a /en/latest/ prefix, so those requests 404 and the 404 page renders unstyled.

Fixed via mkdocs's built-in !ENV tag, which picks up Read the Docs' READTHEDOCS_CANONICAL_URL at build time and falls back to a static default locally. No new dependency needed.


aldbr (changes requested):
LGTM

Since there is no attached issue and Definition of Done here, could you take the opportunity to replace `diracx.io` everywhere?
I still see 2 occurrences of it in the documentation.

Thanks! 🙂

kjvbrt:
Yes, it's done now.

kjvbrt:
PR #1027 needs to be merged first.

aldbr (approved):

